PR #21, opened 2023-07-14
19 defence prompt validation character limit

gsproston-scottlogic:


backend/openai.js
```diff
@@ -99,6 +100,21 @@ async function chatGptChatCompletion() {
 }
 
 async function chatGptSendMessage(message) {
+  // keep track of any triggered defences
+  const defenceInfo = { blocked: false, triggeredDefences: [] };
+  // check if the message is too long
+  if (message.length > 280) {
```
heatherlogan-scottlogic, line 106: do we potentially want to have things like this configurable somewhere?

gsproston-scottlogic: Yeah that's a good idea! I'll make it an environment variable.

gsproston-scottlogic: Just pushed a fix for this.

PR #55, opened 2023-07-24
5 internal documents mvp

heatherlogan-scottlogic:


.env.example
```diff
@@ -13,3 +13,5 @@ SYSTEM_ROLE="Your role is to assist the user with work-related tasks, such as se
 EMAIL_WHITELIST=kate@hotmail.com,bob@hotmail.com,@scottlogic.com
 # Session secret 
 SESSION_SECRET=YOUR_SESSION_SECRET
+# Path to document folder
+DOCUMENT_FOLDER=backend/resources/documents
```
gsproston-scottlogic, line 17: I think because this is unlikely to need to be configured, it would be better to have this as a `const` in the code rather than an environment variable.

backend/src/openai.js
```diff
@@ -89,9 +103,18 @@ async function chatGptCallFunction(functionCall, defenceInfo, session) {
           session
         );
       }
+
     } else if (functionName == "getEmailWhitelist") {
```
gsproston-scottlogic, line 107: I know this isn't new code but we should use `===` here rather than `==`.

gsproston-scottlogic (changes requested):
Looks great! Just some small changes from me.

scottrowan (approved):
all good, I would generally use arrow functions for non react functional components and I think you're a bit overboard with some of the comments, but I can address these things in Typescript migration

PR #63, opened 2023-07-25
Backend integration tests

gsproston-scottlogic:
Planning on adding more integration tests once we've converted to TS.

### DIFF
--- a/backend/src/openai.js
+++ b/backend/src/openai.js
@@ -121,14 +121,18 @@ async function chatGptChatCompletion(session) {
     );
   }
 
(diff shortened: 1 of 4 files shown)


scottrowan (approved):
robust start, especially like how to name your tests, if not sometimes just a bit long-winded,

although same comment as Heather, you really don't need comments on things that are otherwise self-explanatory, e.g. 

```
// initialise OpenAI
initOpenAi();
```

heatherlogan-scottlogic (approved):

PR #71, opened 2023-07-26
Add backend tests to CI

gsproston-scottlogic:


heatherlogan-scottlogic (approved):

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,31 @@
+# This workflow will do a clean installation of node dependencies, cache/restore them, build the source code and run tests across different versions of node
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs
+
+name: Node.js CI
+
+on: [push, pull_request]
```
scottrowan, line 6: I've always also stated the branch, ie dev or main in this case, unless we're happy to run CI on every push to every branch? depends on how long it is I guess

scottrowan (approved):

.github/workflows/node.js.yml
gsproston-scottlogic: Yeah I don't see a problem with just testing on pushes/PRs to main and dev, I'll change that now

Issue #60, opened 2023-07-24
Typescript - Frontend


gsproston-scottlogic:
Refactor the frontend to use Typescript.

PR #77, opened 2023-07-27
migrate frontend to typescript (#60)

scottrowan:
moving to typescript:

major things,
- .js, .jsx files now .ts, .tsx
- tsconfig.json added to root dir
- we never pass "props" always a typed object with a defined interface
- interfaces defined and xported from the service files
- components changed to pass down typed objects rather than individual fields
- all function args and local variables / useStates etc have types defined

sorry the diff isn't so easy to read because its technically deleting / creating new files, not much I can do about that

all my type interfaces are named OpenAI*, such as OpenAIMessage, OpenAIResponse, OpenAIDefence etc, feel free to propose better names


scottrowan:
noticed this will almost certain clash with George's PR, sure merge that first then I'll make a few changes here

frontend/src/components/ChatBox/ChatBox.tsx
```diff
@@ -1,26 +1,36 @@
...
-function ChatBox(props) {
-  const [isSendingMessage, setIsSendingMessage] = useState(false);
-  const [messages, setMessages] = useState([]);
+function ChatBox(
+  this: any,
+  {
+    setEmails,
+    updateTriggeredDefences,
+  }: {
+    setEmails: (emails: any) => void;
```
gsproston-scottlogic, line 18: Could we type emails here with `OpenAIEmail`?

frontend/src/components/DefenceBox/DefenceBox.tsx
```diff
@@ -0,0 +1,99 @@
...
+function DefenceBox({ triggeredDefences }: { triggeredDefences: string[] }) {
+  // list of defence mechanisms
+  const [defenceDetails, setDefenceDetails] = useState(
+    DEFENCE_DETAILS.map((defence) => {
+      return {
+        ...defence,
+        isActive: false,
+        isTriggered: false,
+      };
+    })
```
gsproston-scottlogic, line 21: If we have `isActive` and `isTriggered` default to `false` in the interface definition, we don't need this `map` here.

gsproston-scottlogic (changes requested):
Looks good! I do agree that we should change any references to 'OpenAI' to something else. I reckon if it's related to the LLM, then change it to `chatBot...`, and just remove it from the email and defence interfaces.

frontend/src/service/defenceService.ts
```diff
@@ -0,0 +1,91 @@
+import { sendRequest } from "./backendService";
+
+const PATH = "defence/";
+
+interface OpenAIDefence {
+  name: string;
+  type: string;
+  info: string;
+  isActive?: boolean;
+  isTriggered?: boolean;
```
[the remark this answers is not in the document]
gsproston-scottlogic: Ended up using a class with a constructor here as it doesn't look like it's possible to have default values in a TS interface.

gsproston-scottlogic (approved):

heatherlogan-scottlogic (approved):

PR #85, opened 2023-08-01
80 configure email whitelist defence

heatherlogan-scottlogic:


backend/src/email.js
```diff
@@ -1,18 +1,27 @@
+const { getEmailWhitelistVar } = require("./defence.js");
+const { isDefenceActive } = require("./defence");
```
gsproston-scottlogic, line 2: We should merge these two imports as they're coming from the same file

gsproston-scottlogic (changes requested):
Just the one change from me

gsproston-scottlogic:
Should merge this after #76 is merged

gsproston-scottlogic:
Covered by #90

PR #88, opened 2023-08-03
57 user input api key

heatherlogan-scottlogic:


heatherlogan-scottlogic:
Requires ts-frontend to be merged

frontend/src/components/ApiKeyBox/ApiKeyBox.tsx
```diff
@@ -0,0 +1,76 @@
...
+        id="api-key-input"
+        className={`api-key-input ${isValidated ? "validated" : ""} ${
+          isInvalidated ? "invalidated" : ""
+        }`}
+        type="text"
+        value={apiKey}
+        placeholder="enter your API key here.."
+        onChange={handleApiKeyChange}
+        onKeyUp={handleApiKeySubmit.bind(this)}
+      />
```
gsproston-scottlogic, line 63: Would it be possible to hide the key in some way? I'm thinking like a password input. Just so the user doesn't have their API key on display all the time.

backend/src/openai.js
```diff
@@ -36,29 +36,61 @@ const chatGptFunctions = [
...
+}
+
+function initOpenAi(session) {
   configuration = new Configuration({
-    apiKey: process.env.OPENAI_API_KEY,
+    apiKey: session.apiKey,
   });
   openai = new OpenAIApi(configuration);
+  console.debug("OpenAI initialised");
 }
```
gsproston-scottlogic, line 94: Could we have it so if the .env contains an OPENAI_API_KEY, then it uses that by default, and doesn't require the user to input a key. Would be easier for dev work.

frontend/src/service/openaiService.ts
gsproston-scottlogic, line 1: This file got renamed to 'chatService.ts' so it looks like GitHub thinks this is a new file.

gsproston-scottlogic (changes requested):

gsproston-scottlogic (approved):
Looks great!

Issue #81, opened 2023-07-28
Configure - System role


gsproston-scottlogic:
Allow the user to configure the system role.

Issue #80, opened 2023-07-28
Configure - Email whitelist defence


gsproston-scottlogic:
Allow the user to set the email whitelist.

Issue #79, opened 2023-07-28
Configure - Random sequence enclosure


gsproston-scottlogic:
Include configuration for the random sequence enclosure defence mechanism. This will include the pre-prompt and the sequence length.

Issue #73, opened 2023-07-27
Configure - Character limit defence


gsproston-scottlogic:
Allow the user to configure the character limit.

PR #90, opened 2023-08-03
Configurable defences

gsproston-scottlogic:
PR covering #73 #79 #80 #81


heatherlogan-scottlogic (changes requested):
Looks good to me - but thinking could we have the default values pre-populated inside the input boxes rather than greyed out so the user could select and edit the default values? Thinking it would be easier for the longer prompt fields than having to start from scratch. 
Also if the input boxes could be larger to show more of the prompts

heatherlogan-scottlogic (approved):

PR #120, opened 2023-08-11
98 phase switching

heatherlogan-scottlogic:


backend/src/app.js
```diff
@@ -61,6 +61,9 @@ app.use(function (req, res, next) {
   if (!req.session.gptModel) {
     req.session.gptModel = defaultModel;
   }
+  if (!req.session.phases) {
+    req.session.phases = getPhases();
+  }
```
gsproston-scottlogic, line 66: As discussed we should only need to store one variable here, I'm thinking calling it something like `numPhasesCompleted` and initialising it to `0`. Then we can go up to `3`, indicating that all three phases have been completed by the user.

backend/src/phase.js
gsproston-scottlogic, line 1: Probably don't need this file as the backend will only know about the number of phases the user has completed.

backend/src/router.js
```diff
@@ -168,5 +174,34 @@ router.get("/openai/model", (req, res, next) => {
...
+router.post("/phase/completed", (req, res, next) => {
+  const phase = req.body?.phase;
+  if (phase) {
+    req.session.phases = setPhaseCompleted(req.session, phase);
+    res.send("Phase complete status set");
+  } else {
+    res.statusCode = 400;
+    res.send("Missing phase");
+  }
+});
```
gsproston-scottlogic, line 204: just keep the `router.get("/phase/completed", (req, res, next) => {` endpoint, so the frontend can know which phases the user can switch to. The backend is in charge of completing phases, and the frontend will send the current phase up with the chat message.

frontend/src/models/phase.ts
```diff
@@ -0,0 +1,35 @@
...
+    this.isComplete = isComplete;
+  }
+  name: PHASE_NAMES;
+  index: number;
+  info: string;
+  preamble: string;
+  path_to_documents: string;
+  isCurrent: boolean;
+  isComplete: boolean;
+}
```
gsproston-scottlogic, line 34: I think we could make this an `interface` and export it with `export type`. This keeps the compiled JS nice and clean.

frontend/src/models/phase.ts
```diff
@@ -0,0 +1,35 @@
...
+    this.isCurrent = isCurrent;
+    this.isComplete = isComplete;
+  }
+  name: PHASE_NAMES;
+  index: number;
+  info: string;
+  preamble: string;
+  path_to_documents: string;
+  isCurrent: boolean;
+  isComplete: boolean;
```
gsproston-scottlogic, line 33: The frontend won't be bothered about some of these fields, I'd change it to something like 
```
{
  id: PHASE_NAMES; // unique identifier
  name: string; // display name
  preamble: string;
  isCurrent: boolean;
  isComplete: boolean;
}
```

frontend/src/components/PhaseSelectionBox/PhaseSelectionBox.tsx
```diff
@@ -0,0 +1,80 @@
...
+  getCurrentPhase,
+  setCurrentPhase,
+  setCompletedPhase,
+} from "../../service/phaseService";
+
+import "./PhaseSelectionBox.css";
+
+function PhaseSelectionBox(this: any) {
+  // start on sandbox mode
+  const [currPhase, setCurrPhase] = useState<Phase>(PHASES[3]);
```
gsproston-scottlogic, line 14: Change this to just be a number representing the index to the current phase. eg. `0` for phase 0, `3` for sandbox.

frontend/src/components/PhaseSelectionBox/PhaseSelectionBox.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  // valid phases that the user can go to from the current phase
+  const [currValidPhases, setCurrValidPhases] = useState<number[]>([0]);
```
gsproston-scottlogic, line 17: Change this to be just a number representing the number of completed phases. This is the same as what's stored in the session on the backend, so the frontend can get this from the backend. `0` would mean no phases are completed, so only phase 0 and sandbox are selectable.

frontend/src/components/PhaseSelectionBox/PhaseSelectionBox.tsx
```diff
@@ -0,0 +1,80 @@
...
+    // user can go from sandbox to phase 0 at any time
+    if (phase.index === 3) {
+      return [0, 3];
+      // if current phase is complete, allow user to go to next phase
+    } else if (phase.isComplete) {
+      return [phase.index + 1, 3];
+    }
+    // always allow sandbox mode to be chosen
+    else return [3];
+  };
```
gsproston-scottlogic, line 40: Should be able to use the `isComplete` field to see what's selectable.

gsproston-scottlogic (approved):

PR #131, opened 2023-08-14
115 remove email whitelist defence

gsproston-scottlogic:
In the frontend, I commented out the email whitelist defence from the list of defences, but it is still present in the defence enum.

In the backend, I commented out the `getEmailWhitelist` function so the LLM doesn't think there is a whitelist. 

Code is commented out rather than removed in case we fancy adding it back in later.


heatherlogan-scottlogic (approved):
👍

PR #137, opened 2023-08-14
Not clearing preamble messages

gsproston-scottlogic:


gsproston-scottlogic:
@heatherlogan-scottlogic I've just thought, after this change, the 'clear' button is now more of a 'reset' button. I personally can't think of a scenario where we'd want to clear chat messages and not reset the phase.

heatherlogan-scottlogic (approved):

PR #145, opened 2023-08-15
109 phase 1 win condition

gsproston-scottlogic:
Renamed project X to project XYZ just in case we need to check for the project name in the future. Checking for 'X' might give false positives.

### DIFF
--- a/backend/resources/documents/phase_0/projects.txt
+++ b/backend/resources/documents/phase_0/projects.txt
@@ -1,5 +1,5 @@
 The companies ongoing projects are:
-Project X. 
+Project XYZ. 
 Project EAST. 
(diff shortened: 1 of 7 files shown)


heatherlogan-scottlogic (commented):
Looks great although do we need the project name as well as the manager for the first level? 
Thinking it's an extra level of difficulty if we're trying to make it weak

gsproston-scottlogic:
> Looks great although do we need the project name as well as the manager for the first level? Thinking it's an extra level of difficulty if we're trying to make it weak

ah yeah good point, I think it's enough to email the name.

heatherlogan-scottlogic (approved):

PR #152, opened 2023-08-16
114 phase 2 preamble

heatherlogan-scottlogic:


gsproston-scottlogic (commented):
Might be worth explicitly telling the user that they can toggle and configure defences and how they can do that.

PR #156, opened 2023-08-16
109 phase 1 win condition

gsproston-scottlogic:
Showing a reduced set of attacks in phase 1, as not all are relevant to bypassing a system role.


heatherlogan-scottlogic (approved):
this looks great!

PR #159, opened 2023-08-16
Info message in what when a defence is triggered

heatherlogan-scottlogic:


gsproston-scottlogic (commented):
Looks good, but I'm getting an 'X' even when just toggling defences, is that right?

image

heatherlogan-scottlogic:
> Looks good, but I'm getting an 'X' even when just toggling defences, is that right?
> 
> image

Ah no - fixing

gsproston-scottlogic (approved):

PR #163, opened 2023-08-17
Ts backend

gsproston-scottlogic:


backend/src/app.ts
```diff
@@ -0,0 +1,99 @@
...
+app.use("/", router);
+app.listen(port, () => {
+  console.log("Server is running on port: " + port);
+
+  // for dev purposes only - set the API key from the environment variable
+  const envOpenAiKey = process.env.OPENAI_API_KEY;
+  const prePrompt = getQALLMprePrompt(getInitialDefences());
+  if (envOpenAiKey && prePrompt) {
+    console.debug("Initializing models with API key from environment variable");
+    setOpenAiApiKey(envOpenAiKey, defaultModel, prePrompt);
```
heatherlogan-scottlogic, line 97: I think the secure LLM preprompt is being passed in here by default

heatherlogan-scottlogic (changes requested):
Just one comment as it looks like the QA LLM prompt defence is being activated by default, otherwise looks good!

backend/src/app.ts
gsproston-scottlogic: ah yes, sorting that now

heatherlogan-scottlogic (approved):

PR #165, opened 2023-08-17
confirmation message when defence is configured

heatherlogan-scottlogic:


gsproston-scottlogic (changes requested):
I like this! Although I think the text needs to disappear after a while. For me 'defence successfully configured' sticks around.

heatherlogan-scottlogic:
> I like this! Although I think the text needs to disappear after a while. For me 'defence successfully configured' sticks around.

good point!

gsproston-scottlogic (approved):
Perfect!

PR #171, opened 2023-08-22
91 export log

heatherlogan-scottlogic:


gsproston-scottlogic (commented):
This looks absolutely amazing! Very impressed with this. The only thing I want to ask about is, is there a reason we've got these `const styles =` bits? Or are we ok to make these into .css files?

heatherlogan-scottlogic:
> This looks absolutely amazing! Very impressed with this. The only thing I want to ask about is, is there a reason we've got these `const styles =` bits? Or are we ok to make these into .css files?

Aw thanks it was more tedious than i expected haha! So i believe this react-pdf/renderer package requires these styleSheets rather than importing css for these Page/View types etc

gsproston-scottlogic:
> > This looks absolutely amazing! Very impressed with this. The only thing I want to ask about is, is there a reason we've got these `const styles =` bits? Or are we ok to make these into .css files?
> 
> Aw thanks it was more tedious than i expected haha! So i believe this react-pdf/renderer package requires these styleSheets rather than importing css for these Page/View types etc

Ah right, that makes sense, thanks

PR #177, opened 2023-08-22
turn off configurations for phase 2

heatherlogan-scottlogic:


frontend/src/components/DefenceBox/DefenceMechanism.tsx
```diff
@@ -7,11 +7,13 @@ const { TiTick, TiTimes } = require("react-icons/ti");
 
 function DefenceMechanism({
   defenceDetail,
+  currentPhase,
   setDefenceActive,
   setDefenceInactive,
   setDefenceConfiguration,
 }: {
   defenceDetail: DefenceInfo;
+  currentPhase: number;
```
gsproston-scottlogic, line 16: I'd personally prefer this to be a boolean called something like `showConfiguration`. That way the component doesn't need to know about phases.

gsproston-scottlogic (changes requested):
This works for me, though I think it'd be better if we passed down a boolean for showing defences rather than the phase number.

frontend/src/components/DefenceBox/DefenceMechanism.tsx
heatherlogan-scottlogic: sure can do that

gsproston-scottlogic (approved):
Perfect thank you!

PR #182, opened 2023-08-23
begin integration tests for langchain

heatherlogan-scottlogic:


backend/package.json
```diff
@@ -8,7 +8,8 @@
     "langchain": "^0.0.113",
     "openai": "^3.3.0",
     "pdf-parse": "^1.1.1",
-    "react": "^18.2.0"
+    "react": "^18.2.0",
+    "sinon": "^15.2.0"
```
gsproston-scottlogic, line 12: I've not heard of this, where is it used?

gsproston-scottlogic (commented):
Works great for me! Just a question on an added package.

backend/package.json
heatherlogan-scottlogic: Oh that was a mocking library i tried but didn't use - removing now!

gsproston-scottlogic (approved):

PR #191, opened 2023-08-24
44 defence filtering

heatherlogan-scottlogic:


backend/src/openai.ts
```diff
@@ -320,6 +325,29 @@ async function chatGptSendMessage(
...
+      currentPhase === PHASE_NAMES.SANDBOX
+    ) {
+      const detectedPhrases = detectFilterList(
+        reply.content,
+        getFilterList(defences, DEFENCE_TYPES.FILTER_BOT_OUTPUT)
+      );
+      if (detectedPhrases.length > 0) {
+        console.debug(
+          "FILTER_USER_OUTPUT defence triggered. Detected phrases from blocklist: " +
+            detectedPhrases.join(", ")
```
gsproston-scottlogic, line 341: Some quotes around here too please 🙏

backend/src/router.ts
```diff
@@ -181,6 +195,11 @@ router.post("/openai/chat", async (req, res) => {
             chatResponse.defenceInfo.isBlocked =
               chatResponse.defenceInfo.isBlocked ||
               openAiReply.defenceInfo.isBlocked;
+
+            // combine blocked reason
+            chatResponse.defenceInfo.blockedReason =
+              chatResponse.defenceInfo.blockedReason ||
+              openAiReply.defenceInfo.blockedReason;
```
gsproston-scottlogic, line 202: This is interesting, does this combine both strings?

frontend/src/Defences.ts
```diff
@@ -14,10 +14,16 @@ const DEFENCE_DETAILS_PHASE: DefenceInfo[] = [
   //   [new DefenceConfig("whitelist", "email whitelist")]
   // ),
   new DefenceInfo(
-    DEFENCE_TYPES.LLM_EVALUATION,
-    "LLM Evaluation",
-    "Use an LLM to evaluate the user input for malicious content or prompt injection. ",
-    []
+    DEFENCE_TYPES.FILTER_USER_INPUT,
+    "Input Filtering",
+    "Use a block list of words or phrases to check against user input. If a match is found, the message is block.",
```
gsproston-scottlogic, line 19: Wee typo here, should be `blocked` at the end.

backend/src/defence.ts
```diff
@@ -144,6 +186,22 @@ function generateRandomString(string_length: number) {
   return random_string;
 }
 
+// check message for any words in the filter list
+function detectFilterList(message: string, filterList: string) {
+  const detectedPhrases = [];
+  const cleanedMessage = message.replace(/[^a-zA-Z ]/g, "").toLowerCase();
```
gsproston-scottlogic, line 192: I think it would be good to filter on any string, including ones containing symbols and numbers if possible

backend/src/router.ts
heatherlogan-scottlogic: should be if the user input was blocked, it will show that message, otherwise if the bot response was blocked it will show that message.

gsproston-scottlogic: ah right! Yeah that works for me 👍

gsproston-scottlogic:
There seems to be an issue when adding a string which is just numbers e.g. '999'

gsproston-scottlogic (approved):
Works great thank you 😄

PR #195, opened 2023-08-24
193 UI general

gsproston-scottlogic:
First draft of the UI. From here, we can start the other UI tickets. All colours have been put into the 'Theme.css' file, so they can easily be changed later on if need be.


gsproston-scottlogic:
Here's how it looks for now 

image

heatherlogan-scottlogic (approved):
looks brilliant!

PR #198, opened 2023-08-25
189 UI right side bar

gsproston-scottlogic:


heatherlogan-scottlogic (commented):
theres no "sent emails" header for me now - is that right?

gsproston-scottlogic:
> theres no "sent emails" header for me now - is that right?

That's right, the figma designs don't have that header in.

heatherlogan-scottlogic:
> > theres no "sent emails" header for me now - is that right?
> 
> That's right, the figma designs don't have that header in.

ah neither it does - my bad!

heatherlogan-scottlogic (approved):

PR #201, opened 2023-08-25
WIP: 192 persist chat history for each phase

heatherlogan-scottlogic:


backend/src/router.ts
```diff
@@ -80,41 +91,67 @@ router.post("/defence/configure", (req, res) => {
...
   }
 });
 
 // reset the active defences
 router.post("/defence/reset", (req, res) => {
-  req.session.defences = getInitialDefences();
-  console.debug("Defences reset");
-  res.send("Defences reset");
+  const phase: PHASE_NAMES = req.body?.phase;
+  if (phase >= 0) {
```
gsproston-scottlogic, line 112: Could we just have `if (phase) {` here? And other places where we check for `if (phase >= 0) {`

frontend/src/components/PhaseSelectionBox/PhaseSelectionBox.tsx
```diff
@@ -34,9 +34,7 @@ function PhaseSelectionBox(
               }`}
               key={phase.name}
               onClick={() => handlePhaseChange(index)}
-              disabled={
-                index !== numCompletedPhases && index !== PHASE_NAMES.SANDBOX
-              }
+              disabled={index > numCompletedPhases && index !== 3}
```
gsproston-scottlogic, line 37: Could we change `3` to `PHASE_NAMES.SANDBOX`

gsproston-scottlogic (changes requested):
Looks great! Some very small things from me, and I was running into this bug here where it got a bit confused about the preamble:
image

But this is a chunk of work, well done 🎉

backend/src/router.ts
heatherlogan-scottlogic: i did have this but it returns false when phase = 0 and gives a 400 

image

gsproston-scottlogic: ahh I see. In other parts of the code we've got 
```
  const phase: number | undefined = req.query?.phase as number | undefined;
  if (phase) {
  ```
  Would that work?

heatherlogan-scottlogic: yeah it seem to be fine when inside the get request url 

image

heatherlogan-scottlogic:
> Looks great! Some very small things from me, and I was running into this bug here where it got a bit confused about the preamble: image
> 
> But this is a chunk of work, well done 🎉

Ah thanks! *think* this has fixed the preamble thing now - it kept resurfacing lol

PR #231, opened 2023-09-05
225 multi user langchain

heatherlogan-scottlogic:


backend/src/langchain.ts
```diff
@@ -73,51 +74,77 @@ async function getDocuments(filePath: string) {
...
+      const vectorStore: MemoryVectorStore =
+        await MemoryVectorStore.fromDocuments(documents, embeddings);
+      // store the document vectors for the phase
+      docVectors.push({
+        phase: phase,
+        docVector: vectorStore,
+      });
+    }
+  }
+  setVectorisedDocuments(docVectors);
```
gsproston-scottlogic, line 112: I'm wondering here, `vectorisedDocuments` isn't per user, but embeds a user's API key. Are we able to vectorise the docs  without embedding the key? And maybe embed it later when the user sets their key?

backend/src/router.ts
```diff
@@ -18,14 +16,13 @@ import {
...
 import { chatGptSendMessage, setOpenAiApiKey, setGptModel } from "./openai";
-import { retrievalQAPrePrompt } from "./promptTemplates";
 import { PHASE_NAMES } from "./models/phase";
 import * as fs from "fs";
 
 const router = express.Router();
 
 // keep track of phase change to reinitialize models
-let prevPhase: PHASE_NAMES = PHASE_NAMES.SANDBOX;
+// let prevPhase: PHASE_NAMES = PHASE_NAMES.SANDBOX;
```
gsproston-scottlogic, line 25: I'd say remove this if we don't need it anymore

gsproston-scottlogic (changes requested):

gsproston-scottlogic (changes requested):

PR #236, opened 2023-09-06
Remember defences between phase 2 and sandbox

gsproston-scottlogic:


heatherlogan-scottlogic (changes requested):
looks good just one thing is the info messages in the chat box are printing with underscores like "qa_llm_instructions defence activated"

gsproston-scottlogic:
> looks good just one thing is the info messages in the chat box are printing with underscores like "qa_llm_instructions defence activated"

ah completely missed that

heatherlogan-scottlogic (approved):

PR #245, opened 2023-09-07
237 UI header icon

gsproston-scottlogic:


heatherlogan-scottlogic (changes requested):
The added smile when you win a phase is so good hahah. I saw it moves a bit to the right on some screen sizes, but other than that love it 

image

gsproston-scottlogic:
@heatherlogan-scottlogic I've 'fixed' this by some slightly better css formatting, and not wrapping the app name. It's not ideal but it'll look better when we come up with a (hopefully) shorter app name.

heatherlogan-scottlogic (approved):
Looks great!

PR #247, opened 2023-09-08
241 support more characters in exported logs

heatherlogan-scottlogic:


gsproston-scottlogic:
Just making a comment so we don't forget: we might have to review licenses around certain fonts, to check if we're allowed to use them.

gsproston-scottlogic:
After investigation: the react-pdf/renderer might support multiple fonts, but crucially does not support font _**fallback**_. This unfortunately means it's not currently possible to render all possible languages using this package. 

Relevant issue: https://github.com/diegomura/react-pdf/issues/933

Some possible options are:

1. Get user to select the language when they export.
2. Use a different PDF renderer. 
3. Simply output the log in a different format which simply uses unicode characters and not a renderer.
4. See if we can leverage this PR https://github.com/diegomura/react-pdf/pull/1833

gsproston-scottlogic (approved):

Issue #300, opened 2023-09-19
Streamline frontend refresh network calls 🛠🛠🛠


gsproston-scottlogic:
When the user refreshes the frontend, it makes ~10 API calls to the backend to set up stuff like what model is being used, what state the phases are in, retrieving chat, email, and defences etc. Can we make this one call, where the frontend asks about the current state of the app when the frontend is refreshed. 

<hr>

Here's what happens if I refresh on level 1:

- `health/` always returns with 200 (if the backend is up).
- `email/get?level=0` returns all the sent emails in the session for given level.
- `openai/history?level=0` returns the chat history in the session for given level.
- `defence/status?level=0` returns the defences as configured for given level.
- `openai/validModels` returns all the available models from openAI for the api key configured in the backend.
- `systemRoles/` returns the default system role values for levels 1, 2 and 3.

If we refresh on sandbox we get an additional 4 calls to `openai/model`, all of which return the currently selected model and its config values.

🔧 These should be combined.
🔧 Stop the additional calls to `openai/model` on sandbox.

# AC:

GIVEN frontend in preview mode (so that react doesn't do its double render thing) `npm run preview`
WHEN the user refreshes the website on any level
THEN there should be only one api call made of type fetch

Here's how you quickly find out what api calls are being made that are fetch type. Note that the CombinedFonts call is of type xhr, therefore we ignore it.

![image](https://github.com/ScottLogic/prompt-injection/assets/118171430/48b15e2b-5cbe-4e1c-9495-8848d400c03c)

pmarsh-scottlogic:
investigated under #792 

Here's what happens if I refresh on level 1:
- `health/` always returns with 200 (if the backend is up).
- `email/get?level=0` returns all the sent emails in the session for given level.
- `openai/history?level=0` returns the chat history in the session for given level.
- `defence/status?level=0` returns the defences as configured for given level.
- `openai/validModels` returns all the available models from openAI for the api key configured in the backend.
- `systemRoles/` returns the default system role values for levels 1, 2 and 3.

pmarsh-scottlogic:
- add a GET levelState endpoint which returns
  - emails
  - chat history
  - configured defences
- Remove all calls and related code to the health endpoint from the frontend.

PR #659, opened 2023-12-13
634 frontend hangs when it doesnt get a chat reply

gsproston-scottlogic:
## Description

In the event that a chat request to the backend times out, the frontend now shows and error and can recover without refreshing.

## Screenshots

**Scenario 1: Backend isn't running at start**

Backend is started again after the first reply.

image

**Scenario 2: Backend is running to begin with**

Backend is stopped after the first reply.
Backend is started again after the third reply.

image

## Notes

- I removed a bit of code that would get the level's emails when the ChatBox component is mounted. As far as I could see, this wasn't doing anything as emails are passed to the ChatBox from MainComponent through MainBody,
- I also updated the way we get newly sent emails. Now, emails that have been sent due to a user chat message are sent back as part of the chat reply object. This reduces the number of network calls as before we were sending a request to get all sent emails after getting a chat reply.

## Concerns

- I was tempted to combine the two existing networks calls to get the level's chat history and emails into one call, something like `getLevelState`. I decided against it as I thought it might be covered better in #300 .

## Checklist

Have you done the following?

- [x] Added tests


chriswilty:
@gsproston-scottlogic The single test failure is that sporadic one that I commented out, so you might want to rebase on latest dev and try again...

backend/test/unit/controller/chatController.test.ts
```diff
@@ -68,6 +74,30 @@ describe('handleChatToGPT unit tests', () => {
...
+										confirmed: true,
+									}),
+								},
+							},
+						],
+					},
+				},
+			],
+		};
+	}
```
chriswilty, line 99: "prettier" my ass!!! Gotta laugh...

chriswilty (approved):
Beautiful 👌

PR #662, opened 2023-12-13
646 chat history exceeds models max context length

heatherlogan-scottlogic:
## Description

Addressing the bug where the chat history sent to the model exceeded the max number of tokens, so it threw an error. It was down to our calculations being off few so the filterChatHistoryByMaxTokens function did not cut down the history enough. 

## Screenshots

## Notes
- the promptTokensEstimate function from the openai-chat-tokens should calculate tokens for whole chat history plus initial function definitions etc. As it does not support tool_calls yet, there was a discrepancy of a few tokens. have added in a manual calculation for the tool calls to append to this total. We can verify the manual count is correct from the logs: The first prompt_tokens comes from our calculation, the second is from the response sent from OpenAI.
image

- changed the way we do filterChatHistoryByMaxTokens , before we would loop through the list and calculate the tokens of an individual message using messageTokensEstimate function openai-chat-tokens, and add these until the sum exceeded the max tokens. however there were differences in the counts given by summing the messageTokensEstimate outputs, compared to the promptTokensEstimate output. So instead i've changed it to add a message one by one to a list, calculate the whole total tokens using promptTokensEstimate , and remove the last message if it doesn't fit. so the token estimate by this function will match that of openai. 
- both of these changes should ensure that the original issue where more tokens than before were sent to gpt and the error occurred. 
- I also added a 5% discount to the max tokens per model (see line 319 of openai.ts) .this gives us some wiggle room if the number begins to approach the max, and also makes sure the model has some tokens to respond with in that case. 
- refactored some openAI types (e.g ChatCompletionMessage -> ChatCompletionMessageParam) also required a chatResponse.reply.toString() as the new completion reply type was string | ChatCompletionPart.

- when we filter the list, if we remove a message with tool_calls, we have to remove the corresponding assistant reply that has the same tool_call_id. Otherwise we get this error (which is what line 100-104 of tokens.ts is for) 
image

## Concerns

* left in some logs with prompt_tokens= as i found it useful to keep track although ik it's maybe cumbersome 
🤐
## Checklist

Have you done the following?

- [x] Added tests


backend/src/openai.ts
```diff
@@ -32,9 +30,13 @@ import {
...
 	FunctionSendEmailParams,
 } from './models/openai';
+import chatModelMaxTokens, {
+	countTotalPromptTokens,
+	filterChatHistoryByMaxTokens,
+} from './utils/token';
 
 // tools available to chatGPT
-const chatGptTools: ChatCompletionTool[] = [
+export const chatGptTools: ChatCompletionTool[] = [
```
gsproston-scottlogic, line 39: Prefer this export to be at the end of the file rather than inline. Unfortunately I don't _think_ there's a linting rule for this.

backend/src/utils/token.ts
```diff
@@ -0,0 +1,115 @@
...
+		estimatedTokens,
+		'maxNumTokens=',
+		maxNumTokens
+	);
+	const newList: ChatCompletionMessageParam[] = [];
+
+	// reverse list to add from most recent
+	const reverseHistory = chatHistory.slice().reverse();
+	// always add first message
+	newList.push(reverseHistory[0]);
```
gsproston-scottlogic, line 76: I'm sure the `filterChatHistoryByMaxTokens` method is not called if there's no chat history, but I think it's still worth checking that the `chatHistory` length is > 0 in this method.

backend/src/utils/token.ts
```diff
@@ -0,0 +1,115 @@
...
+				newList.splice(i - 1, 1);
+			}
+			break;
+		}
+	}
+	return newList.reverse();
+}
+
+export {
+	chatModelMaxTokens as default,
```
gsproston-scottlogic, line 112: I'd prefer not to export as default here.

gsproston-scottlogic (changes requested):
Great stuff, only very minor requests from me.

Incidentally, How accurate is the token counting method for the tool calls? Is it worth perhaps opening a PR on that package or are we just guestimating here?

backend/src/controller/chatController.ts
```diff
@@ -34,7 +34,7 @@ async function handleLowLevelChat(
 		req.session.levelState[currentLevel].sentEmails,
 		currentLevel
 	);
-	chatResponse.reply = openAiReply.completion?.content ?? '';
+	chatResponse.reply = openAiReply.completion?.content?.toString() ?? '';
```
pmarsh-scottlogic, line 37: How does this behave when `openAiReply.completion.content` is of type `ChatCompletionContentPart[]`? Or is that not expected to happen? Couldn't find much online about it

backend/src/openai.ts
```diff
@@ -302,97 +298,6 @@ async function chatGptChatCompletion(
...
-			sumTokens += numTokens;
-		} else {
-			console.debug('Max tokens reached on completion = ', element);
-			break;
-		}
-	}
-	return filteredList.reverse();
-}
-
 // take only the completions to send to GPT
```
pmarsh-scottlogic, line 302: Not really in PR scope, but I think this comment and the one on line 306 don't need to be there (I think they don't give any info beyond what's in the method name)

backend/src/utils/token.ts
pmarsh-scottlogic, line 1: Love that this is in a new file now

backend/src/utils/token.ts
```diff
@@ -0,0 +1,115 @@
...
+		if (message.role === 'assistant' && message.tool_calls) {
+			numToolCalls += message.tool_calls.length;
+			tokens += countSingleToolCallTokens(message.tool_calls);
+		}
+	}
+	return tokens + numToolCalls * 4; // 4 tokens per tool call
+}
+
+// estimate total number of tokens across whole chat history
+function countTotalPromptTokens(chatHistory: ChatCompletionMessageParam[]) {
```
pmarsh-scottlogic, line 44: I'd prefer to remove comment and call method `estimateTotalPromptTokens`

backend/src/utils/token.ts
```diff
@@ -0,0 +1,115 @@
...
+function countToolCallTokens(chatHistory: ChatCompletionMessageParam[]) {
+	let numToolCalls = 0;
+	let tokens = 0;
+	for (const message of chatHistory) {
+		if (message.role === 'assistant' && message.tool_calls) {
+			numToolCalls += message.tool_calls.length;
+			tokens += countSingleToolCallTokens(message.tool_calls);
+		}
+	}
+	return tokens + numToolCalls * 4; // 4 tokens per tool call
```
pmarsh-scottlogic, line 40: particularly pernickety suggestion from me
```suggestion
        const TOKENS_PER_TOOL_CALL = 4;
	return tokens + numToolCalls * TOKENS_PER_TOOL_CALL ;
```

backend/src/utils/token.ts
```diff
@@ -0,0 +1,115 @@
...
+	return toolCall.reduce((acc, toolCall) => {
+		const {
+			function: { name, arguments: args },
+		} = toolCall;
+		return acc + stringTokens(name) + stringTokens(args);
+	}, 0);
+}
+
+// count total tool call in chat histroy tokens - not supported by openai-chat-tokens yet. To be removed when supported
+function countToolCallTokens(chatHistory: ChatCompletionMessageParam[]) {
```
pmarsh-scottlogic, line 31: I don't agree with the comment. We'll still need this method that adds the tool calls across the chat history! Therefore I suggest the comment be removed.

Also I'd prefer the method to be named `estimateTOtalToolCallTokens`

backend/src/utils/token.ts
```diff
@@ -0,0 +1,115 @@
+import {
+	ChatCompletionMessageParam,
+	ChatCompletionMessageToolCall,
+} from 'openai/resources/chat/completions';
+import { promptTokensEstimate, stringTokens } from 'openai-chat-tokens';
+
+import { CHAT_MODELS } from '@src/models/chat';
+import { chatGptTools } from '@src/openai';
+
+// max tokens each model can use
```
pmarsh-scottlogic, line 10: I think the const name explains this pretty well on its own

backend/src/utils/token.ts
```diff
@@ -0,0 +1,115 @@
...
+		promptTokensEstimate({
+			messages: chatHistory,
+			functions: chatGptTools.map((tool) => tool.function),
+		}) +
+		countToolCallTokens(chatHistory) +
+		5
+	); // there is an offset of 5 between openai completion prompt_tokens
+}
+
+// filter older messages out of chat history to fit inside max tokens a model can use
```
pmarsh-scottlogic, line 55: Don't think the comment is necessary given the good method name

backend/test/unit/utils/token.test.ts
```diff
@@ -0,0 +1,304 @@
+import { ChatCompletionMessageParam } from 'openai/resources/chat/completions';
+
+import { filterChatHistoryByMaxTokens } from '@src/utils/token';
+
+describe('token unit tests', () => {
+	test('GIVEN chat history exceeds max token number WHEN applying filter THEN it should return the filtered chat history', () => {
+		const maxTokens = 260; // 120 tokens for function call
```
pmarsh-scottlogic, line 7: this comment confuses me

backend/test/unit/utils/token.test.ts
```diff
@@ -0,0 +1,304 @@
...
+		];
+		const filteredChatHistory = filterChatHistoryByMaxTokens(
+			chatHistory,
+			maxTokens
+		);
+		expect(filteredChatHistory).toEqual(chatHistory);
+	});
+
+	test('GIVEN the max tokens is reached on a tool call WHEN applying filter THEN it should remove the previous assistant reply to the tool call', () => {
+		const maxTokens = 200; // 120 tokens for function call
```
pmarsh-scottlogic, line 250: same here. What is this comment trying to say?

pmarsh-scottlogic (changes requested):
Some relatively minor and nitpicky comments from me. In general it's about the use of comments. My view (which I acknowledge might be a bit extreme) is the only reason you should use a comment is to explain "why" something is rather than "what" something is. If you need a comment to explain what something is or is doing, then the naming in the code could be improved.

But that's just my opinion and I'm happy to be disagreed with :)

backend/src/utils/token.ts
heatherlogan-scottlogic: when the openai-chat-tokens library supports tool calls then they will be counted inside promptTokensEstimate - so this method will need to be removed otherwise we'll be counting them twice :)

Also can I ask why you prefer estimate? I was using countFunction to distinguish between our own functions and the estimateFunctions provided by the library

backend/test/unit/utils/token.test.ts
heatherlogan-scottlogic: the model is set up with function call definitions (or that is chatGptTools) which uses up 120 tokens before any thing is added to the chat history. 
So if maxTokens = 260, 120 are already used for the function calls, then the rest can be used in chat history. 

I will update the comments to make this clearer.

backend/src/controller/chatController.ts
heatherlogan-scottlogic: my understanding of this is that this would only be used in cases where we can input images and text - as we can't do that it should only ever by a string :) 

These are the types 

export type ChatCompletionMessageParam =
  | ChatCompletionSystemMessageParam
  | ChatCompletionUserMessageParam
  | ChatCompletionAssistantMessageParam
  | ChatCompletionToolMessageParam
  | ChatCompletionFunctionMessageParam;

all of those ..MessageParams have content: string except this user input ones

export interface ChatCompletionUserMessageParam {
  /**

   * The contents of the user message.
   */
  content: string | Array<ChatCompletionContentPart> | null;

  /**
   * The role of the messages author, in this case `user`.
   */
  role: 'user';
}

which is for image input

backend/src/utils/token.ts
pmarsh-scottlogic: Ahhh my bad - I misunderstood on both counts. Well justified. I guess I like estimate because it makes it clear that it's not "official", it's conjecture. But I agree it's more important to differentiate it from library methods.
No change necessary then :)

backend/test/unit/utils/token.test.ts
pmarsh-scottlogic: ah, I see!

backend/src/controller/chatController.ts
pmarsh-scottlogic: thank you for explaining this to me. Gosh, this API is complicated!

backend/src/utils/token.ts
```diff
@@ -37,28 +39,28 @@ function countToolCallTokens(chatHistory: ChatCompletionMessageParam[]) {
...
 function countTotalPromptTokens(chatHistory: ChatCompletionMessageParam[]) {
 	return (
 		promptTokensEstimate({
 			messages: chatHistory,
 			functions: chatGptTools.map((tool) => tool.function),
 		}) +
 		countToolCallTokens(chatHistory) +
-		5
-	); // there is an offset of 5 between openai completion prompt_tokens
+		OFFSET_OPENAI_TOKENS
```
pmarsh-scottlogic, line 52: Ah, this is nice too!

backend/src/utils/token.ts
```diff
@@ -37,28 +39,28 @@ function countToolCallTokens(chatHistory: ChatCompletionMessageParam[]) {
...
 }
 
-// filter older messages out of chat history to fit inside max tokens a model can use
 function filterChatHistoryByMaxTokens(
 	chatHistory: ChatCompletionMessageParam[],
 	maxNumTokens: number
 ): ChatCompletionMessageParam[] {
+	if (chatHistory.length === 0) {
+		return chatHistory;
+	}
```
pmarsh-scottlogic, line 62: nice easy bit of optimisation!

backend/test/unit/utils/token.test.ts
```diff
@@ -3,8 +3,11 @@ import { ChatCompletionMessageParam } from 'openai/resources/chat/completions';
 import { filterChatHistoryByMaxTokens } from '@src/utils/token';
 
 describe('token unit tests', () => {
+	// model will be set up with function definitions so will contribute to maxTokens
+	const FUNCTION_DEF_TOKENS = 120;
+
 	test('GIVEN chat history exceeds max token number WHEN applying filter THEN it should return the filtered chat history', () => {
-		const maxTokens = 260; // 120 tokens for function call
+		const maxTokens = FUNCTION_DEF_TOKENS + 140;
```
pmarsh-scottlogic, line 10: I'm a fan of this

pmarsh-scottlogic (approved):
Cracking stuff

gsproston-scottlogic (approved):
Nice!

PR #681, opened 2023-12-18
676 switched input and export/reset buttons

dhinrichs-scottlogic:
## Description
Reversed the order of 
- input, send button
and 
- export chat,  reset level buttons

## Screenshot Outcome:
Image

## Additional context
This is helpful for keyboard-only users (going between the chat and the configurations for example), assistive tech, and means the export chat & reset buttons stay in place even as the chat input expands with larger input. 

## Acceptance criteria
**GIVEN** the user chats with the chatbot
**WHEN** they tab through the interactive elements
**THEN** the input and send button come before the export chat and reset level buttons.


pmarsh-scottlogic (approved):
✨

heatherlogan-scottlogic (approved):
👌

PR #682, opened 2023-12-18
561 update UI when reset triggered

heatherlogan-scottlogic:
## Description

Adds a message to the chat history when user clicks "reset level" button (no message is shown for the reset progress button)

## Screenshots

image

## Notes

- added new info message type
- changed the logic in adding a welcome message to ensure reset message appears first (was trickier as we do not add the welcome message to the chat history on backend, so need to sort the list when setting the front end messages) 
- also noticed resetting does not turn deactivate the sandbox so fixed that - I think we would want to deactivate the active defences but keep the configured values when we reset. 

## Concerns

- I followed the top answer [here ](https://stackoverflow.com/questions/5214127/css-technique-for-a-horizontal-line-with-words-in-the-middle) to style the horizontal line in the reset message, where we have a line and wrap the text in a span with white background over the top. Do you know if this would have any accessibility implications @dhinrichs-scottlogic 
- 
- <s> in the ticket we decided not to disable the reset level button if there's nothing to clear. however I've added an ableToReset function to stop the request being sent in that case (to stop spamming the button), but this is essentially the same as disabling button.. Accessibility wise - should we allow the button to be clickable (if it's having no effect), and if so do we need designs to make the disabled button visually different from the export chat button?  </s>
- ** Update Removed the ableToReset check after talking to @dhinrichs-scottlogic about this button issue - decided it's best to have the reset message still appear as visual feedback to the user 

## Checklist

Have you done the following?

- [ ] Added tests


pmarsh-scottlogic (approved):
Looks grand!

pmarsh-scottlogic:
I don't think there's any need disabling the button visually or functionally when there's nothing to reset, but if you want to save processing by having abeToReset, then that's cool.

dhinrichs-scottlogic:
A few thoughts about whether or not to disable the buttons:
- disabling buttons is not generally recommended if it can be avoided
- When you _can_ click a button, you expect something to happen when you click the button, therefore if possible, can we display the "Level Progress reset" message even without sending the request to the backend?

Issue #620, opened 2023-11-29
New Email UI


gsproston-scottlogic:
# Feature Request

## Description
Implement the new UI designs for emails.

## Screenshots
![image](https://github.com/ScottLogic/prompt-injection/assets/103250539/bb26e0c8-c256-41a1-ac68-7ec8e06da1be)

## Additional context
[Figma](https://www.figma.com/file/dntI3xrEI40De9KDCMCmMI/AI-prompt-injection?type=design&node-id=335-1931&mode=dev)

No functionality change here, just UI update.

## Acceptance criteria
Ensure that colour contrast passes accessibility standards.
Firefox is good at this. There's an accessibility tab in dev tools where you can see colour contrast warnings.

**WHEN** the user sends an email
**THEN** the email is shown in the right panel with the correct content

dhinrichs-scottlogic:
Found a bug!

**Description:**
Email outbox section doesn't scroll anymore, isn't keyboard focusable, and there is no padding in between emails.

**Steps to reproduce the behaviour:**
1. Go to any level
2. Ask the chatbot to send multiple emails, until the email outbox section on the right hand side fills up completely
3. See the emails squished together in the outbox container

**Expected behaviour**
When the emails are too big, or too many to fit into their container,
the user can scroll through the container to see all emails. 
There is padding in between each email, and the email container is keyboard
focusable and can be scrolled by keyboard and mouse. 

**Screenshot of issue**
![Image](https://github.com/ScottLogic/prompt-injection/assets/125262707/d31b77bc-dbb0-4ec0-aaf1-53b860653533)

heatherlogan-scottlogic:
testing

PR #688, opened 2023-12-19
620-email-ui-fix

pmarsh-scottlogic:
## Description

Makes emails behave properly when they fill up the sent emails container. [Bug found in testing](https://github.com/ScottLogic/prompt-injection/issues/620#issuecomment-1860018216).

## Screenshots

image

## Notes

- Fixed it by removing `overflow: hidden`. However, I couldn't keep `the border-radius` property without the overflow property. It would cause the rounded corners on the outer div, bu the inner divs wouldn't get clipped, and they would remain with their radius-0 corners, so no visual change. So I've had to apply the `border-radius` directly to the child elements.

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/frontend/src/components/EmailBox/EmailBox.css
+++ b/frontend/src/components/EmailBox/EmailBox.css
@@ -6,4 +6,5 @@
 	overflow-y: auto;
 	padding: 1rem;
 	scrollbar-width: thin;
+	gap: 1rem;
 }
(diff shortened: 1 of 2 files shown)


dhinrichs-scottlogic (approved):
works well, thank you!

gsproston-scottlogic (approved):
Good stuff! Sorry looks like it was my border radius change that caused this :')

PR #689, opened 2023-12-19
677 group configurations in fieldset where possible

heatherlogan-scottlogic:
## Screenshots

UI appearance is the same. 
fieldset grouping

image

## Notes

- Wrapped ModelConfigurationSlider in <fieldset>
- changes` <p>config.name</p> to <legend>`
- some styling

## Concerns

* I found that i couldn't position the legend to appear on the same line as the button (it would always take up a line to itself - think this is by design as found others complaining about it online?) - workaround was to use a negative margin-top on the button to get on the same line - not ideal but looks like it looks fine on different size screens.. 

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/frontend/src/components/ModelBox/ModelConfigurationSlider.css
+++ b/frontend/src/components/ModelBox/ModelConfigurationSlider.css
@@ -2,23 +2,22 @@ fieldset {
 	border: none;
 	margin: 0;
 	padding: 0;
-}
-
-.model-config-info {
 	display: flex;
-	align-items: center;
-	padding: 0.5rem 0;
+	flex-direction: column;
 }
 
-.model-config-info legend {
+legend {
 	margin-right: auto;
 }
 
-.model-config-info-icon {
+.info-icon {
 	font-size: 1.3rem;
+	margin-left: auto;
+	margin-top: -1.3rem;
+	margin-bottom: 1rem;
 }
 
-.model-config-info-text {
+.info-text {
 	font-size: 0.875rem;
 }
 
--- a/frontend/src/components/ModelBox/ModelConfigurationSlider.css
+++ b/frontend/src/components/ModelBox/ModelConfigurationSlider.css
@@ -1,4 +1,4 @@
-fieldset {
+.model-config-slider-fieldset {
 	border: none;
 	margin: 0;
 	padding: 0;
@@ -6,7 +6,7 @@ fieldset {
 	flex-direction: column;
 }
 
-legend {
+.model-config-slider-fieldset legend {
 	margin-right: auto;
 }
 
@@ -24,7 +24,7 @@ legend {
 .model-config-slider {
 	width: 90%;
 	margin: 0 auto;
-	padding: 0 0 0 0;
+	padding: 0;
 	display: flex;
 	flex-direction: column;
 	justify-content: center;
--- a/frontend/src/components/ModelBox/ModelConfigurationSlider.css
+++ b/frontend/src/components/ModelBox/ModelConfigurationSlider.css
@@ -11,10 +11,7 @@
 .model-config-info {
 	display: flex;
 	align-items: center;
-}
-
-.model-config-title {
-	margin-right: auto;
+	padding-bottom: 0.5rem;
 }
 
 .model-config-info-icon {
@@ -23,4 +20,13 @@
 
 .model-config-info-text {
 	font-size: 0.875rem;
+}
+
+fieldset {
+	border: 0;
+	padding: 0;
+}
+
+legend {
+	margin-right: auto;
 }
--- a/frontend/src/components/ModelBox/ModelConfigurationSlider.css
+++ b/frontend/src/components/ModelBox/ModelConfigurationSlider.css
@@ -10,18 +10,18 @@
 	margin-right: auto;
 }
 
-.info-icon {
+.model-config-slider-fieldset .info-icon {
 	font-size: 1.3rem;
 	margin-left: auto;
 	margin-top: -1.3rem;
 	margin-bottom: 1rem;
 }
 
-.info-text {
+.model-config-slider-fieldset .info-text {
 	font-size: 0.875rem;
 }
 
-.model-config-slider {
+.model-config-slider-fieldset .config-slider {
 	width: 90%;
 	margin: 0 auto;
 	padding: 0;
--- a/frontend/src/components/ModelBox/ModelConfigurationSlider.css
+++ b/frontend/src/components/ModelBox/ModelConfigurationSlider.css
@@ -1,17 +1,17 @@
-.model-config-slider {
-	width: 90%;
-	margin: 0 auto;
-	padding: 0 0 0 0;
-	display: flex;
-	flex-direction: column;
-	justify-content: center;
-	align-items: center;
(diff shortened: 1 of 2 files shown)


frontend/src/components/ModelBox/ModelConfigurationSlider.tsx
```diff
@@ -32,11 +32,9 @@ function ModelConfigurationSlider({
 	const supportText = `more info about ${config.name}`;
 
 	return (
-		<div>
+		<fieldset>
 			<div className="model-config-info">
```
dhinrichs-scottlogic, line 36: can we switch the `<div className="model-config-info>` and  `<legend>` around so that the legend comes right after the fieldset? that way the fieldset has a label. That would require some css adjustments as well to move the icon again in the right place. 
You can check the accessibility tab in firefox (right click, second option from below, "Inspect Accessibility Properties")  and look in the accessibility tree if the fieldset (shows up as "group" in the accessibility tree) has been labelled correctly.

frontend/src/components/ModelBox/ModelConfigurationSlider.css
```diff
@@ -1,26 +1,32 @@
-.model-config-slider {
-	width: 90%;
-	margin: 0 auto;
-	padding: 0 0 0 0;
+fieldset {
```
gsproston-scottlogic, line 1: This will apply this styling to all over fieldsets, which is problematic if we want to add them to other components. So instead I'd rather have a class name, and prepend all other sections in this style sheet with that class name

frontend/src/components/ModelBox/ModelConfigurationSlider.css
```diff
@@ -1,26 +1,32 @@
...
-	align-items: center;
-}
-
-.model-config-info {
-	display: flex;
-	align-items: center;
 }
 
-.model-config-title {
+legend {
```
gsproston-scottlogic, line 9: So if the class name for the `fieldset` is "model-slider-fieldset", then this becomes `.model-slider-fieldset legend`, for example.

frontend/src/components/ModelBox/ModelConfigurationSlider.css
```diff
@@ -1,26 +1,32 @@
...
 
-.model-config-info-text {
+.info-text {
 	font-size: 0.875rem;
 }
+
+.model-config-slider {
+	width: 90%;
+	margin: 0 auto;
+	padding: 0 0 0 0;
```
gsproston-scottlogic, line 27: can just be `padding: 0` ?

gsproston-scottlogic (changes requested):
Neat! Just a naming thing from me.

pmarsh-scottlogic (changes requested):
Nothing in the code beyond what George has already spotted.

I notice the AC says this:
"THEN the interactive elements are grouped and labels aren't repeated unnecessarily"

But when you tab through the controls, NVDA says things like:
- "model temperature grouping more info about model temperature button"
- "model temperature grouping temperature slider"
- "Top P grouping more info about top p button"
- "Top P grouping Top P slider"

... etc

which sounds like unnecessary repetition to me. But I guess it's not an entirely objective criterion.
Anyway, if we don't like how it is then I see two solutions:
- make the `aria-labels` more generic
- remove the fieldset. Personally I have reservations about it - all the examples I've seen group multiple input elements together within a form context, which I can see being useful. In this case I can see it's applying a description to the section which includes the info button and slider, but those elements have very descriptive labels anyway.

Sorry - it's a bit late for me to oppose the ticket. But I'm happy to keep it as long as we don't think the screenreader output is too repetitive.

@dhinrichs-scottlogic what do you think of the screen reader output? Does the fieldset help other assistive technologies in different ways?

frontend/src/components/ModelBox/ModelConfigurationSlider.tsx
```diff
@@ -32,23 +32,19 @@ function ModelConfigurationSlider({
...
-				>
-					<AiOutlineInfoCircle aria-hidden />
-				</button>
-			</div>
-			{showInfo && <div className="model-config-info-text">{config.info}</div>}
+		<fieldset className="model-config-slider-fieldset">
+			<legend>{config.name}</legend>
+			<button
+				className="info-icon prompt-injection-min-button"
+				title={supportText}
```
dhinrichs-scottlogic, line 39: I'd say remove the supportText const and change the title to "more info" as well? 
Interestingly, if the title and aria-label are different, they will both be read out, whereas when they are the same, at least NVDA doesn't repeat them, which is neat

frontend/src/components/ModelBox/ModelConfigurationSlider.tsx
```diff
@@ -29,15 +29,13 @@ function ModelConfigurationSlider({
...
 	return (
-		<fieldset>
+		<fieldset className="model-config-slider-fieldset">
 			<legend>{config.name}</legend>
 			<button
 				className="info-icon prompt-injection-min-button"
-				title={supportText}
-				aria-label={supportText}
+				title="more info"
+				aria-label="more info"
```
pmarsh-scottlogic, line 38: just a tiny preference: when I was screen reading, having "info" read out by a robot felt kinda weird, but that's just my opinion. I'd prefer "information", but I guess this is cool if it's more consistent with the tone of the rest of the app.

pmarsh-scottlogic (approved):
🚀

frontend/src/components/ModelBox/ModelConfigurationSlider.tsx
dhinrichs-scottlogic: saves 2 vowels of time though!

dhinrichs-scottlogic (approved):

frontend/src/components/ModelBox/ModelConfigurationSlider.tsx
pmarsh-scottlogic: Didn't consider that!

frontend/src/components/ModelBox/ModelConfigurationSlider.css
```diff
@@ -1,26 +1,32 @@
...
-	align-items: center;
 }
 
-.model-config-title {
+.model-config-slider-fieldset legend {
 	margin-right: auto;
 }
 
-.model-config-info-icon {
+.info-icon {
```
gsproston-scottlogic, line 13: Could we have `.model-config-slider-fieldset` before this too? 👀

frontend/src/components/ModelBox/ModelConfigurationSlider.css
```diff
@@ -1,26 +1,32 @@
...
 	font-size: 1.3rem;
+	margin-left: auto;
+	margin-top: -1.3rem;
+	margin-bottom: 1rem;
 }
 
-.model-config-info-text {
+.info-text {
```
gsproston-scottlogic, line 20: And this please 🙏 I just think they're quite general class names otherwise

gsproston-scottlogic (changes requested):
Sorry just two more small things.

gsproston-scottlogic (approved):
Perfect thanks 🙏

PR #692, opened 2023-12-19
571 update chat styles bubbles

gsproston-scottlogic:
## Description

Updated the style for the transformed user messages.

## Screenshots

Normal XML tagging:

image

Escaping XML tags:

image

## Notes

- What was happening before was the chat response would contain a single string for the transformed message. In the event that the message was not transformed, that field would still be set, but it would just be the exact same as the user's message. Now, if the message is not transformed, that field is not set in the response.
- Split the transformed message into three parts: the actual message, and the bits before and after it. This is so we could **bold** the actual message for clarity.

## Concerns

- <s>I noticed while doing this that we're appending the original untransformed message to the chat history, rather than the transformed one. Do we still want to do this?</s> **- we actually are adding the transformed message to the chat history. The comment is describing adding the info message, so it can be retrieved on the frontend.**

## Checklist

Have you done the following?

- [x] Added tests


backend/src/controller/chatController.ts
```diff
@@ -48,19 +53,20 @@ async function handleHigherLevelChat(
...
-	const messageIsTransformed = chatResponse.transformedMessage !== message;
-	if (messageIsTransformed) {
+	if (transformedMessage) {
+		chatResponse.transformedMessage = transformedMessage;
+		// if message has been transformed then add the original to chat history and send transformed to chatGPT
 		req.session.levelState[currentLevel].chatHistory.push({
 			completion: null,
 			chatMessageType: CHAT_MESSAGE_TYPE.USER,
 			infoMessage: message,
 		});
```
gsproston-scottlogic, line 67: I've kept the existing functionality here... but I'm thinking we should add the transformed message to the chat history. What do we think? 🤔

heatherlogan-scottlogic: Ah so i've worded that comment badly here (was confused myself for a while), i meant that the original message is added as an infoMessage and not a completion, which would mean it does not get sent to openAI but can still access it in the frontend chat. 
The transformed message is still added to the chatHistory as a completion, so it will be the one that is sent to openAI 

image

gsproston-scottlogic: ahh I see, thanks for looking into that

heatherlogan-scottlogic (approved):
all looks good to me, splitting up message is clever 🤓

pmarsh-scottlogic (approved):
Really like the transformedMessage in the response being null if there was no transformation. Much more readable!

Cracking job.

PR #693, opened 2023-12-19
675 chat messages should show up in chronological order

dhinrichs-scottlogic:
# Feature Request

## Description
Currently, the new messages show up as the "oldest" element in the DOM-tree. This means that if you navigate the chatbox with assistive tech, you have to skip through from the oldest messages until you reach the newest message. 
We should find a way to reverse this order. 
The same is true for the email outbox. 
It should also be clear who is speaking for each message (chatbot, or user) 

## Acceptance criteria

**GIVEN** You navigate the chatbox with assistive tech
**WHEN** you navigate through the chat messages and the email outbox
**THEN** you get the messages in correct chronological order and know the sender (and recipient for email) 

## Notes

- the chatbox container and each message has turned into a section element, and are not keyboard focusable, which makes it easier to go through the messages and re-read them one by one with assistive tech. 
- aria-live has been added to the chatbox section and email outbox section, so that updates are announced automatically (works better on narrator than NVDA)
- each message now has a visually hidden span that announces the author/type (e.g. "Chatbot said", "Information message") of the message
- The order of the messages and email has been reversed, and a useEffect hook to autoscroll the container added. 
- since the keyboard-focusable messages add many focusable elements, hidden level 2 headings have been added to email outbox, chatbot section, and configurations to create landmarks that are easy to skip to. 

## Concerns
- My experience of using screen readers (Narrator and NVDA) is limited, and different screen readers and browsers work differently with aria-live, so I'm sure this can be improved on, but aiming for progress, not perfection. 

## Checklist

Have you done the following?

- [ ] Added tests


frontend/src/components/ChatBox/ChatBoxFeed.tsx
```diff
@@ -12,14 +13,21 @@ function ChatBoxFeed({ messages }: { messages: ChatMessage[] }) {
...
+		}
+	}, [messages]);
+
 	return (
-		<div
+		<section
 			className="chat-box-feed"
 			ref={chatboxFeedContainer}
-			// eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
 			tabIndex={isOverflow ? 0 : undefined}
```
gsproston-scottlogic, line 27: Now that messages inside the feed are focusable, we don't need this anymore. So we can also remove the `/* eslint-disable jsx-a11y/no-noninteractive-tabindex */` at the top of this file.

frontend/src/components/ChatBox/ChatBoxInfoText.tsx
```diff
@@ -11,8 +12,17 @@ function ChatBoxInfoText({
...
 	type: CHAT_MESSAGE_TYPE;
 }) {
+	const messageType =
+		type === CHAT_MESSAGE_TYPE.INFO
+			? 'Information message '
+			: type === CHAT_MESSAGE_TYPE.DEFENCE_ALERTED
+			? 'Information message '
+			: type === CHAT_MESSAGE_TYPE.DEFENCE_TRIGGERED
+			? 'Information message '
+			: 'unknown message type ';
```
gsproston-scottlogic, line 22: Can simplify this bit. We have three conditions that return the same value.

frontend/src/components/EmailBox/EmailBox.tsx
```diff
@@ -12,16 +12,18 @@ function EmailBox({ emails }: { emails: EmailInfo[] }) {
 	const isOverflow = useIsOverflow(emailBoxContainer);
 
 	return (
-		<div
+		<section
 			className="email-box"
 			ref={emailBoxContainer}
 			// eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
 			tabIndex={isOverflow ? 0 : undefined}
```
gsproston-scottlogic, line 19: Similarly here we shouldn't need this anymore as individual emails are now focusable.

frontend/src/components/ChatBox/ChatBoxFeed.tsx
```diff
@@ -12,14 +13,21 @@ function ChatBoxFeed({ messages }: { messages: ChatMessage[] }) {
 	const chatboxFeedContainer = useRef<HTMLDivElement>(null);
 	const isOverflow = useIsOverflow(chatboxFeedContainer);
 
+	useEffect(() => {
+		if (chatboxFeedContainer.current) {
+			chatboxFeedContainer.current.scrollTop =
+				chatboxFeedContainer.current.scrollHeight;
+		}
+	}, [messages]);
```
gsproston-scottlogic, line 21: We'll also need one of these for the EmailBox

gsproston-scottlogic (changes requested):
Looks good, just a few things from me.

frontend/src/components/ChatBox/ChatBoxFeed.tsx
```diff
@@ -10,16 +9,21 @@ import './ChatBoxFeed.css';
...
 
 	return (
-		<div
+		<section
 			className="chat-box-feed"
 			ref={chatboxFeedContainer}
-			// eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
-			tabIndex={isOverflow ? 0 : undefined}
+			aria-live="polite"
```
pmarsh-scottlogic, line 24: I'm amused by this.
I'm guessing it means that we don't want updates to the feet to interrupt whatever's being said?

frontend/src/components/ChatBox/ChatBoxInfoText.tsx
```diff
@@ -11,8 +12,16 @@ function ChatBoxInfoText({
 	text: string;
 	type: CHAT_MESSAGE_TYPE;
 }) {
+	const messageType =
+		type === CHAT_MESSAGE_TYPE.INFO ||
+		type === CHAT_MESSAGE_TYPE.RESET_LEVEL ||
+		type === CHAT_MESSAGE_TYPE.DEFENCE_ALERTED ||
+		type === CHAT_MESSAGE_TYPE.DEFENCE_TRIGGERED
```
pmarsh-scottlogic, line 19:
```suggestion
		type === CHAT_MESSAGE_TYPE.INFO ||
		type === CHAT_MESSAGE_TYPE.RESET_LEVEL ||
		type === CHAT_MESSAGE_TYPE.DEFENCE_ALERTED ||
		type === CHAT_MESSAGE_TYPE.DEFENCE_TRIGGERED ||
		type === CHAT_MESSAGE_TYPE.RESET_LEVEL
```

frontend/src/components/ChatBox/ChatBoxMessage/MessageBubble.tsx
```diff
@@ -34,16 +34,33 @@ function MessageBubble({
...
+			: message.type === CHAT_MESSAGE_TYPE.USER
+			? 'You said: '
+			: message.type === CHAT_MESSAGE_TYPE.USER_TRANSFORMED
+			? 'Your message transformed by XML tagging: '
+			: message.type === CHAT_MESSAGE_TYPE.ERROR_MSG
+			? 'Error message: '
+			: message.type === CHAT_MESSAGE_TYPE.BOT
+			? 'ScottBrewBot said:  '
+			: 'ScottBrewBot said: ';
+
```
pmarsh-scottlogic, line 49: this is really good

frontend/src/components/ControlPanel/ControlPanel.tsx
```diff
@@ -57,6 +57,7 @@ function ControlPanel({
 			{(currentLevel === LEVEL_NAMES.LEVEL_3 ||
 				currentLevel === LEVEL_NAMES.SANDBOX) && (
 				<>
+					<h2 className="visually-hidden">Chatbot Configurations</h2>
```
pmarsh-scottlogic, line 60: unsure about "Chatbot Configurations". I'd argue the defence configurations are not chatbot configurations. So I guess I'd recommend either just "configurations" or "defence and model configurations"

pmarsh-scottlogic (changes requested):
smashed it.

Just a few minor adjustments from me

I'm still getting the problem where sometimes the screenreader will read the entire contents of the chat feed when I send a message. Happy for that to be a separate ticked though.

frontend/src/components/ChatBox/ChatBoxInfoText.tsx
gsproston-scottlogic: Good catch!

frontend/src/components/ChatBox/ChatBoxMessage/MessageBubble.tsx
```diff
@@ -34,16 +34,33 @@ function MessageBubble({
...
 			)}
 			{message.type === CHAT_MESSAGE_TYPE.LEVEL_INFO && (
 				<p className="level-info-header">Information</p>
 			)}
-			{message.message}
-		</div>
+			<p>
+				<span className="visually-hidden">{messageAuthor}</span>
+				{message.message}
+			</p>
```
gsproston-scottlogic, line 62: This `<p>` element is adding too much spacing to the chat bubbles:

image

Just need to remove the margin.

frontend/src/components/ChatBox/ChatBoxFeed.tsx
gsproston-scottlogic, line 1: It looks like the chat and email feeds are coming down from the top, rather than up from the bottom. 

image

Can we fix that?

frontend/src/components/EmailBox/EmailBox.tsx
gsproston-scottlogic, line 1: When the email feed scrolls, the whole application gets a scrollbar too o.O

image

gsproston-scottlogic (changes requested):
Code looks good, but I'm seeing some visual issues now :/

frontend/src/components/ChatBox/ChatBoxFeed.tsx
dhinrichs-scottlogic: "polite" is the recommended setting. And no, there is no "rude" setting, I checked ;)
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

frontend/src/components/EmailBox/EmailBox.tsx
dhinrichs-scottlogic: oh dear, more work to do! Thanks for your keen eye!

dhinrichs-scottlogic:
@pmarsh-scottlogic 
> I'm still getting the problem where sometimes the screenreader will read the entire contents of the chat feed when I send a message. Happy for that to be a separate ticked though.

Different screen readers (and the same screen reader in two different browsers) use aria-live differently, so results will be inconsistent. 
I've found that NVDA will fairly reliably only read out the new messages, whereas Narrator will repeat the new updates twice in chrome, and not read out any updates at all in Firefox... I haven't found a way to fix that.
Looking for info on how to make chatbots accessible, I read that each chat message should be keyboard-focusable (even though it's not an interactive element), and in a section element. This does help with this issue as well, because you can easily tab-through messages back and forth, so you don't have to wait until the screen reader has announced everything, you can just tab to the most recent message.

frontend/src/components/ChatBox/ChatBoxInfoText.tsx
dhinrichs-scottlogic: Why RESET_LEVEL twice?

frontend/src/components/ControlPanel/ControlPanel.tsx
pmarsh-scottlogic: if we don't want to lump the view documents button with the config, can we give it its own header?

pmarsh-scottlogic:
> @pmarsh-scottlogic
> 
> > I'm still getting the problem where sometimes the screenreader will read the entire contents of the chat feed when I send a message. Happy for that to be a separate ticked though.
> 
> Different screen readers (and the same screen reader in two different browsers) use aria-live differently, so results will be inconsistent. I've found that NVDA will fairly reliably only read out the new messages, whereas Narrator will repeat the new updates twice in chrome, and not read out any updates at all in Firefox... I haven't found a way to fix that. Looking for info on how to make chatbots accessible, I read that each chat message should be keyboard-focusable (even though it's not an interactive element), and in a section element. This does help with this issue as well, because you can easily tab-through messages back and forth, so you don't have to wait until the screen reader has announced everything, you can just tab to the most recent message.

ye the tabbable thing is super convenient. I had a pretty long conversation with my screen off and it wasn't too shabby. Shame about the inconsistencies, but I guess it's kinda out of our ballpark

frontend/src/components/ChatBox/ChatBoxInfoText.tsx
pmarsh-scottlogic: silly me lol
All good, ignore me!

gsproston-scottlogic: yes sorry I can't read 👓

frontend/src/components/ControlPanel/ControlPanel.tsx
dhinrichs-scottlogic: I don't like the idea of having a separate header for the document viewer button, the idea of headers is to create skip links. While the configurations don't change the chatbot itself, they do change how the chatbot interacts in practice with the user. 
How about "Chatbot System Configurations"?

frontend/src/components/ChatBox/ChatBoxMessage/MessageBubble.tsx
```diff
@@ -54,12 +54,13 @@ function MessageBubble({
 				<b>Transformed: </b>
 			)}
 			{message.type === CHAT_MESSAGE_TYPE.LEVEL_INFO && (
-				<p className="level-info-header">Information</p>
+				<>
+					<p className="level-info-header">Information</p>
+					<span className="visually-hidden"> message</span>
```
gsproston-scottlogic: Interesting! How does this work? Does a screen reader not read out the message twice here?

frontend/src/components/ChatBox/ChatBoxMessage/MessageBubble.tsx
```diff
@@ -34,11 +34,29 @@ function MessageBubble({
...
+			? ''
+			: message.type === CHAT_MESSAGE_TYPE.USER
+			? 'You said: '
+			: message.type === CHAT_MESSAGE_TYPE.USER_TRANSFORMED
+			? 'Your message transformed by XML tagging: '
+			: message.type === CHAT_MESSAGE_TYPE.ERROR_MSG
+			? 'Error message: '
+			: message.type === CHAT_MESSAGE_TYPE.BOT
+			? 'ScottBrewBot said:  '
+			: 'ScottBrewBot said: ';
```
gsproston-scottlogic, line 48: Tiny, but seeing as these are almost the same, we could simplify this. Unless the extra space is critical!
```suggestion
			: 'ScottBrewBot said: ';
```

gsproston-scottlogic (changes requested):
Beautiful! Just one very small thing from me

frontend/src/components/ChatBox/ChatBoxMessage/MessageBubble.tsx
gsproston-scottlogic: Checked and it doesn't (at least not on windows narrator), so looks all good!

frontend/src/components/ChatBox/ChatBoxMessage/MessageBubble.tsx
dhinrichs-scottlogic: I thought it was, but turns out it's not!

frontend/src/components/ChatBox/ChatBoxMessage/MessageBubble.tsx
dhinrichs-scottlogic: I changed the message author 
this is just "message" as a string, not the {message}. 
That way, a screen reader will read: 
"Information message
Congratulations" You have completed this level. Please click on the next level to continue".

gsproston-scottlogic (approved):
Excellent 😇

frontend/src/components/ChatBox/ChatBoxMessage/MessageBubble.tsx
gsproston-scottlogic: ahh I see! Thanks for explaining

frontend/src/components/ChatBox/ChatBoxInfoText.css
```diff
@@ -6,7 +6,7 @@
 }
 .chat-box-info.info-text {
 	color: var(--chat-info-text-colour);
-	margin-top: 0.1rem;
+	margin: 0.1rem 0.1rem;
```
pmarsh-scottlogic, line 9:
```suggestion
	margin: 0.1rem;
```
same thing :)

frontend/src/components/ChatBox/ChatBoxMessage/MessageBubble.tsx
```diff
@@ -36,26 +36,25 @@ function MessageBubble({
...
-			)}
+		<section className={className} lang="en" tabIndex={0}>
 			{message.type === CHAT_MESSAGE_TYPE.LEVEL_INFO && (
-				<p className="header">Information</p>
+				<>
+					<p className="header">Information</p>
+					<span className="visually-hidden"> message: </span>
+				</>
 			)}
+			<span className="visually-hidden">{messageAuthor}</span>
```
pmarsh-scottlogic, line 57: This is real pedantic, so you can ignore if you wish, but I'd prefer to only render the messageAuthor span if it's going to have a value.
i.e.
```typeescript
{message.type === CHAT_MESSAGE_TYPE.LEVEL_INFO: (
				<>
					<p className="header">Information</p>
					<span className="visually-hidden"> message: </span>
				</>
			) ? (<span className="visually-hidden">{messageAuthor}</span>)
}
```

frontend/src/components/ControlPanel/ControlPanel.tsx
```diff
@@ -55,7 +55,7 @@ function ControlPanel({
 			{(currentLevel === LEVEL_NAMES.LEVEL_3 ||
 				currentLevel === LEVEL_NAMES.SANDBOX) && (
 				<>
-					<h2 className="visually-hidden">Chatbot Configurations</h2>
+					<h2 className="visually-hidden">ScottBrew System Access</h2>
```
pmarsh-scottlogic, line 58: that's quite a fun title since it fits in with the story!

frontend/src/components/MainComponent/MainBody.css
```diff
@@ -9,16 +9,24 @@
 
 .centre-area {
 	display: flex;
-	flex-direction: column;
+	flex-direction: column-reverse;
```
pmarsh-scottlogic, line 12: I flicked this on and off and I don't believe it's doing anything? Which would make sense given it has only one child

image

pmarsh-scottlogic (changes requested):
Works beautifully! Some very small changes remaining from me

frontend/src/components/MainComponent/MainBody.css
pmarsh-scottlogic: In fact, I've checked the other properties and the same argument applies. This class can simply be
```css
.centre-area {
  width:55%;
  background-color: var(--main-text-colour);
}
```

frontend/src/components/ChatBox/ChatBoxMessage/MessageBubble.tsx
dhinrichs-scottlogic: For consistency and clarity I kind of want to keep it as it is for now if that's okay?

pmarsh-scottlogic: sure thing. I suppose clarity is subjective!

pmarsh-scottlogic (approved):
that's the badger 🦡

PR #698, opened 2023-12-20
598 have themedbutton always use aria disabled over disabled

gsproston-scottlogic:
## Description

The ThemedButton component now only uses `aria-disabled` rather than `disabled`.

## Notes

- Also removed the `appearsDifferentWhenDisabled` from the component.
- Wrapped the LoadingButton component in a new `div` in order to overwrite the disabled styling on the child ThemedButton.
- Actually fixed a bug where the LoadingButton wasn't being disabled at all due to a typo in the `ariaDisabled` prop name (it was `aria-disabled`).
- Simplified the LoadingButton props.
- There was a default label on the spinner inside the LoadingButton that made it quite odd when it was being read out by a screen reader. The label was "three-dots-loading". The screen reader said something along the lines of "send three dots loading button disabled". Which I thought was clunky so I removed the label, so now it says "send button disabled".

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/frontend/src/components/ThemedButtons/LoadingButton.tsx
+++ b/frontend/src/components/ThemedButtons/LoadingButton.tsx
@@ -18,7 +18,13 @@ function LoadingButton({
 			<ThemedButton ariaDisabled={isLoading} onClick={onClick}>
 				{children}
 				{isLoading && (
-					<ThreeDots width="1.5rem" color="white" wrapperClass="loader" />
+					<ThreeDots
+						width="1.5rem"
+						color="white"
+						wrapperClass="loader"
+						// blank label as by default the label is 'three-dots-loading'
+						ariaLabel=""
+					/>
 				)}
 			</ThemedButton>
 		</div>
--- a/frontend/src/components/ThemedButtons/LoadingButton.tsx
+++ b/frontend/src/components/ThemedButtons/LoadingButton.tsx
@@ -14,20 +14,22 @@ function LoadingButton({
 	onClick: () => void;
 }) {
 	return (
-		<div className="loading-button">
-			<ThemedButton ariaDisabled={isLoading} onClick={onClick}>
-				{children}
-				{isLoading && (
-					<ThreeDots
-						width="1.5rem"
-						color="white"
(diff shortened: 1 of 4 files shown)


dhinrichs-scottlogic (approved):
Works well with NVDA and Narrator in Firefox and Chrome! 
Thank you :)

frontend/src/components/ThemedButtons/LoadingButton.tsx
```diff
@@ -1,27 +1,33 @@
...
-			disabled={isLoading}
-			appearsDifferentWhenDisabled={false}
-			{...buttonProps}
-		>
-			{children}
-			{isLoading && (
-				<ThreeDots width="1.5rem" color="white" wrapperClass="loader" />
-			)}
-		</ThemedButton>
+		<div className="loading-button">
```
pmarsh-scottlogic, line 17: not a huge fan of this single-child div. I'd much rather apply styles directly to the themed button via an extra classname

pmarsh-scottlogic (changes requested):
I have but a single qualm

frontend/src/components/ThemedButtons/LoadingButton.tsx
gsproston-scottlogic: hmm yes I would rather get rid of the div, but I wasn't sure how to override the disabled styling on the ThemedButton otherwise. Whereabout do you suggest this extra classname should be?

pmarsh-scottlogic: add it as a prop to themedButton, and within the code for themedButton, append it to the existing classname. Like what I did to the headerButton here https://github.com/ScottLogic/prompt-injection/pull/633/commits/ebd0b042821b4d8d0e07cbaa3055ffe09f0369f0

gsproston-scottlogic: elegant! done now

pmarsh-scottlogic (approved):
yessssss

PR #701, opened 2023-12-20
fix bot blocked position on export log

heatherlogan-scottlogic:
## Description

Missed out BOT_BLOCKED in getMessageStyle case stmt. 🤷‍♂️ Also added a wee Bot (blocked): for clarity:

## Screenshots

image

## Notes

- Bullet point list with
- any notable code changes

## Concerns

- Bullet point list
- with any questions
- or concerns about this PR

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/frontend/src/components/ExportChat/ExportChatMessage.tsx
+++ b/frontend/src/components/ExportChat/ExportChatMessage.tsx
@@ -55,8 +55,9 @@ function getFullPrefix(message: ChatMessage) {
 		case CHAT_MESSAGE_TYPE.ERROR_MSG:
 			return `Error: ${message.message}`;
 		case CHAT_MESSAGE_TYPE.BOT:
-		case CHAT_MESSAGE_TYPE.BOT_BLOCKED:
 			return `Bot: ${message.message}`;
+		case CHAT_MESSAGE_TYPE.BOT_BLOCKED:
+			return `Bot (blocked): ${message.message}`;
 		default:
 			return message.message;
 	}
@@ -71,6 +72,7 @@ function getMessageStyle(type: CHAT_MESSAGE_TYPE) {
 			return styles.chatBoxInfo;
 		case CHAT_MESSAGE_TYPE.USER:
 			return styles.chatBoxMessage;
+		case CHAT_MESSAGE_TYPE.BOT_BLOCKED:
 		case CHAT_MESSAGE_TYPE.BOT:
 			return styles.chatBoxMessageBot;
 		case CHAT_MESSAGE_TYPE.LEVEL_INFO:


gsproston-scottlogic (approved):
hyuge 🙆

frontend/src/components/ExportChat/ExportChatMessage.tsx
```diff
@@ -71,6 +72,7 @@ function getMessageStyle(type: CHAT_MESSAGE_TYPE) {
 			return styles.chatBoxInfo;
 		case CHAT_MESSAGE_TYPE.USER:
 			return styles.chatBoxMessage;
+		case CHAT_MESSAGE_TYPE.BOT_BLOCKED:
```
pmarsh-scottlogic, line 75: I don't quite get what's going on with these empty case statements. How does it behave? Could we write it explicitly?

pmarsh-scottlogic (approved):
good catch :)

one comment but it's not really to do with the ticket so happy to approve

frontend/src/components/ExportChat/ExportChatMessage.tsx
gsproston-scottlogic: Empty case statements just drop down to the next case statement. It's fairly common to see a "fallthrough" comment in these cases, so we could do that.

```typescript
case CHAT_MESSAGE_TYPE.BOT_BLOCKED:
        // fallthrough
case CHAT_MESSAGE_TYPE.BOT:
	return styles.chatBoxMessageBot;
```

pmarsh-scottlogic: so 
```typescript
case CHAT_MESSAGE_TYPE.BOT_BLOCKED:
        // fallthrough
case CHAT_MESSAGE_TYPE.BOT:
	return styles.chatBoxMessageBot;
```
is the same as
```typescript
case CHAT_MESSAGE_TYPE.BOT_BLOCKED || CHAT_MESSAGE_TYPE.BOT:
	return styles.chatBoxMessageBot;
```
?

I guess it's more aesthetically pleasing this way. If this is a common thing to do then I'm happy with it!

gsproston-scottlogic: Sort of, except the latter doesn't quite work. It's sort of the same as doing 
`if (type === (CHAT_MESSAGE_TYPE.BOT_BLOCKED || CHAT_MESSAGE_TYPE.BOT))`
which would be like doing
`if (type === false)`
which isn't what you want.

With switch statements, if you want the same result for multiple values, yes the former is the way to do it 😃

pmarsh-scottlogic: Ahhh, right! Thanks for educating me :)

PR #718, opened 2023-12-22
704 refactor detecttriggereddefences to remove mutation

pmarsh-scottlogic:
## Description

r-r-refactor, y'all 🏭
specifically, killing the mutants 🔫

## Notes

as the ticket says, `detectTriggeredDefences()` now follows this structure:
```typescript
const xmlDefenceRep = detectXmlDefence(...)
const promptEvalRep = detectPromptEvalDefence(...)
...
return combineReports(xmlDefenceRep, promptEvalRep, ...)
```

And each the detection method for each defence no longer mutates anything.

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/backend/src/defence.ts
+++ b/backend/src/defence.ts
@@ -223,6 +223,14 @@ async function detectTriggeredDefences(message: string, defences: Defence[]) {
 
 	return defenceReport;
 }
+function getEmptyChatDefenceReport(): ChatDefenceReport {
+	return {
+		blockedReason: null,
+		isBlocked: false,
+		alertedDefences: [],
+		triggeredDefences: [],
+	};
+}
 
 function detectCharacterLimit(
 	defenceReport: ChatDefenceReport,
@@ -230,7 +238,7 @@ function detectCharacterLimit(
 	defences: Defence[]
 ) {
 	const maxMessageLength = Number(getMaxMessageLength(defences));
-	// check if the message is too long
+
 	if (message.length > maxMessageLength) {
 		console.debug('CHARACTER_LIMIT defence triggered.');
 		// check if the defence is active
@@ -244,8 +252,10 @@ function detectCharacterLimit(
 			// add the defence to the list of alerted defences
 			defenceReport.alertedDefences.push(DEFENCE_ID.CHARACTER_LIMIT);
 		}
+		return defenceReport;
+	} else {
+		return getEmptyChatDefenceReport();
 	}
-	return defenceReport;
 }
 
 function detectFilterUserInput(
--- a/backend/src/defence.ts
+++ b/backend/src/defence.ts
@@ -268,31 +268,35 @@ function detectFilterUserInput(
 		message,
 		getFilterList(defences, DEFENCE_ID.FILTER_USER_INPUT)
 	);
-	if (detectedPhrases.length > 0) {
+
+	const filterWordsDetected = detectedPhrases.length > 0;
+	const defenceActive = isDefenceActive(DEFENCE_ID.FILTER_USER_INPUT, defences);
+
+	if (filterWordsDetected) {
 		console.debug(
 			`FILTER_USER_INPUT defence triggered. Detected phrases from blocklist: ${detectedPhrases.join(
 				', '
 			)}`
 		);
-		if (isDefenceActive(DEFENCE_ID.FILTER_USER_INPUT, defences)) {
-			return {
+	}
+
+	return filterWordsDetected && defenceActive
+		? {
 				blockedReason: `Message blocked - I cannot answer questions about '${detectedPhrases.join(
 					"' or '"
 				)}'!`,
 				isBlocked: true,
 				alertedDefences: [],
 				triggeredDefences: [DEFENCE_ID.FILTER_USER_INPUT],
-			};
-		} else {
-			return {
+		  }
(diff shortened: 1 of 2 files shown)


gsproston-scottlogic (changes requested):
While this all works, I don't actually think each of the detect methods (e.g. `detectCharacterLimit`) each need to return a full defence report.

The reason being is that each of them are only ever going to report a single element in the `alertedDefences` or `triggeredDefences` arrays, and even then never in both of the arrays.

I've also noticed that there is always a `blockedReason` when `isBlocked` is `true`. So we could just remove `isBlocked` and see instead if `blockedReason` is `null` or not.

I'm thinking instead they could return a smaller report, something like 
```typescript
interface singleDefenceReport {
        blockedReason: string | null;
	status: 'altered' | 'triggered' | 'ok'
}
```
so in `detectCharacterLimit` you could have
```typescript
return {
	blockedReason: messageExceedsLimit && defenceActive ? 'Message is too long' : null,
	status: !messageExceedsLimit ? 'ok' : defenceActive ? 'triggered' : 'alerted',
}
```
and then in `detectTriggeredDefences` have running arrays for triggered and alerted defences.
```typescript
const alertedDefences: DEFENCE_ID[] = [];
const triggeredDefences: DEFENCE_ID[] = [];

const characterLimitDefenceReport = detectCharacterLimit(message, defences);
if (characterLimitDefenceReport.status === 'triggered')
	triggeredDefences.push(DEFENCE_ID.CHARACTER_LIMIT);
else if (characterLimitDefenceReport.status === 'alerted')
	alertedDefences.push(DEFENCE_ID.CHARACTER_LIMIT);
```
You can then combine the blocked reasons as you're already doing.

pmarsh-scottlogic:
> While this all works, I don't actually think each of the detect methods (e.g. `detectCharacterLimit`) each need to return a full defence report.
> 
> The reason being is that each of them are only ever going to report a single element in the `alertedDefences` or `triggeredDefences` arrays, and even then never in both of the arrays.
> 
> I've also noticed that there is always a `blockedReason` when `isBlocked` is `true`. So we could just remove `isBlocked` and see instead if `blockedReason` is `null` or not.
> 
> I'm thinking instead they could return a smaller report, something like
> 
> ```ts
> interface singleDefenceReport {
>         blockedReason: string | null;
> 	status: 'altered' | 'triggered' | 'ok'
> }
> ```
> 
> so in `detectCharacterLimit` you could have
> 
> ```ts
> return {
> 	blockedReason: messageExceedsLimit && defenceActive ? 'Message is too long' : null,
> 	status: !messageExceedsLimit ? 'ok' : defenceActive ? 'triggered' : 'alerted',
> }
> ```
> 
> and then in `detectTriggeredDefences` have running arrays for triggered and alerted defences.
> 
> ```ts
> const alertedDefences: DEFENCE_ID[] = [];
> const triggeredDefences: DEFENCE_ID[] = [];
> 
> const characterLimitDefenceReport = detectCharacterLimit(message, defences);
> if (characterLimitDefenceReport.status === 'triggered')
> 	triggeredDefences.push(DEFENCE_ID.CHARACTER_LIMIT);
> else if (characterLimitDefenceReport.status === 'alerted')
> 	alertedDefences.push(DEFENCE_ID.CHARACTER_LIMIT);
> ```
> 
> You can then combine the blocked reasons as you're already doing.

You're right, I was thinking it was quite wasteful, but couldn't think of a better thing. This solution is elegant!

pmarsh-scottlogic:
@gsproston-scottlogic adapted my solution to look more like yours. I wasn't a great fan of checking each SingleDefenceReport in a an if statement to see whether it should be added to the array, so I've kept the old logic of combining it all at the end. All methods are super pure, and we only return the necessary stuff, and I like how it reads. What do you think?

heatherlogan-scottlogic (approved):
lovely refactor ⭐ can't see anything i'd improve upon!

backend/src/defence.ts
```diff
@@ -207,117 +211,149 @@ function transformMessage(
...
+		: null;
 
-	return defenceReport;
+	const alertedDefences = defenceReports
+		.filter((report) => report.status === 'alerted')
+		.map((report) => report.defence);
+
+	const triggeredDefences = defenceReports
+		.filter((report) => report.status === 'triggered')
+		.map((report) => report.defence);
```
gsproston-scottlogic, line 241: Much more elegant than what I proposed! Excellent stuff.

gsproston-scottlogic (approved):
Perfect 👌 excellent work 😎

backend/src/defence.ts
```diff
@@ -207,117 +211,149 @@ function transformMessage(
...
-	await detectEvaluationLLM(defenceReport, message, defences);
+function combineDefenceReports(
+	defenceReports: SingleDefenceReport[]
+): ChatDefenceReport {
+	const isBlocked = defenceReports.some((report) => report.blockedReason);
+	const blockedReason = isBlocked
+		? defenceReports
+				.map((report) => report.blockedReason)
+				.filter((reason) => reason !== null)
+				.join('\n')
```
chriswilty, line 232: Not something you need to change, but just thought I'd point out it's _nanofractionally_ more efficient to filter and then map, rather than map and then filter.

chriswilty (approved):
Like this a lot 👌

PR #719, opened 2023-12-22
678 reset level resets defences but doesnt update UI

gsproston-scottlogic:
## Description

Added a confirmation modal after clicking 'reset level', and also made sure that defences are NOT reset.

## Screenshots

Overlay on levels 1 and 2 (no defences)

image

Overlay on level 3 and sandbox (with defences)

image

Before reset:

image

After reset:

image

## Notes

- Refactored how the overlay works so sub components can set the html for the overlay. This is nice as it means we don't need to move a ton of logic into App.tsx.
- Made a new general overlay component called OverlayChoice, which is used by both ResetLevel and ResetProgress overlays.

## Concerns

- I was tempted to also use OverlayChoice with the welcome overlay and the overlay that shows after beating level 3. However, these actually have some nice aria-labels that are particular to switching modes, so I left them.
- Nothing calls `resetActiveDefences` anymore, but I've left it in for if we want to have a "reset all defences" button. Happy to take that out if people prefer.

## Checklist

Have you done the following?

- [ ] Added tests


frontend/src/App.tsx
pmarsh-scottlogic, line 1: I like the refactor

frontend/src/components/ChatBox/ChatBox.tsx
```diff
@@ -75,6 +71,13 @@ function ChatBox({
 			: `Congratulations, you have completed the final level of your assignment!`;
 	}
 
+	function isLevelComplete() {
+		// level is complete if the chat contains a LEVEL_INFO message
+		return messages.some(
+			(message) => message.type === CHAT_MESSAGE_TYPE.LEVEL_INFO
+		);
+	}
+
```
pmarsh-scottlogic, line 80: very nice! Reads like a book :)

frontend/src/components/MainComponent/MainBody.tsx
```diff
@@ -19,12 +17,12 @@ function MainBody({
...
 	addSentEmails,
 	resetDefenceConfiguration,
-	resetLevel,
 	toggleDefence,
 	setDefenceConfiguration,
 	incrementNumCompletedLevels,
-	openLevelsCompleteOverlay,
 	openDocumentViewer,
+	openLevelsCompleteOverlay,
+	openResetLevel,
```
pmarsh-scottlogic, line 25: nitpick but I'd prefer `openResetLevelOverlay`

pmarsh-scottlogic (approved):
great job with the refactor, I think it reads nice and clearly. I'd prefer to remove that unused method though. Can look back in git history if we want the code back

heatherlogan-scottlogic (approved):
all looks/works good for me! lovely refactor ⭐

PR #720, opened 2023-12-22
703 add random sequence enclosure defence back in

heatherlogan-scottlogic:
## Description

Defence is back in and new radio buttons option to apply XML tagging or RSE

## Screenshots

None selected (sandbox)
image

XML tagging selected (sandbox)
image

random sequence selected (sandbox)
image

level 3 
image

Toggling XML, then toggling random Sequence in chatbox
image

## Notes

- added in random sequence enclosure defence from deleted code, with some small changes
- new defence type PROMPT_ENCLOSURE 
- new frontend component PromptEnclosureDefenceMechanism to display the radio buttons. it is rendered in DefenceMechanism instead of DefenceConfiguration when DEFENCE_ID.PROMPT_ENCLOSURE
- added the defence name to TransformedChatMessage so we can use it in the info messages (thinking - do we want an enum for the defence names, of could we replace the DEFENCE_ID strings to be the display value instead of CHARACTER_LIMIT etc. don't think these are used anywhere?) 

## Concerns

- Was unsure what to name this, went with prompt enclosure 
- I've got an generic info text "Alter the users input by appending instructional prompts and/or tags to the input." for Prompt Enclosure, then a shortened specific line for XML tagging/RSE. do we want to get rid of the generic one?
- Not sure about 'None' option - is it accessible etc?

## Checklist

Have you done the following?

- [x] Added tests


frontend/src/Defences.ts
```diff
@@ -52,9 +52,26 @@ const DEFENCES_SHOWN_LEVEL3: Defence[] = [
...
+		[
+			makeDefenceConfigItem('PROMPT', 'prompt', 'text'),
+			makeDefenceConfigItem('SEQUENCE_LENGTH', 'sequence length', 'number'),
+		]
+	),
+
+	makeDefence(
+		DEFENCE_ID.PROMPT_ENCLOSURE,
+		'Prompt Enclosure',
+		'Alter the users input by appending instructional prompts and/or tags to the input.',
```
gsproston-scottlogic, line 71: "users" should have an apostrophe (lines 55, 61, 71)

frontend/src/components/DefenceBox/PromptEnclosureDefenceMechanism.tsx
```diff
@@ -0,0 +1,106 @@
...
+			if (defence.isActive) {
+				toggleDefence(defence);
+			}
+		});
+	}
+
+	// activate this defence and deactivate all others
+	function activateOne(defenceId: DEFENCE_ID) {
+		defences.forEach((defence) => {
+			if (defence.id === defenceId && !defence.isActive) toggleDefence(defence);
```
gsproston-scottlogic, line 48: Just a personal preference but I'd prefer braces around the `if` statement if the `else` statement has braces.

frontend/src/components/DefenceBox/PromptEnclosureDefenceMechanism.tsx
```diff
@@ -0,0 +1,106 @@
...
+			else if (defence.isActive) {
+				toggleDefence(defence);
+			}
+		});
+	}
+
+	function handleRadioChange(event: React.ChangeEvent<HTMLInputElement>) {
```
gsproston-scottlogic, line 55: I'm wondering if it might be better for this function to take just the string value, because the only time `event` is used here is to get the value using `event.target.value`. 

So this method would become
```typescript
function handleRadioChange(defenceId: string) {
```

frontend/src/components/ThemedButtons/DefenceConfigurationRadioButton.css
```diff
@@ -0,0 +1,30 @@
+.defence-radio-button {
+	position: relative;
+	cursor: pointer;
+	display: inline-block;
+}
+
+.defence-radio-button input {
+	opacity: 0;
```
gsproston-scottlogic, line 8: This is unfortunately also hiding the focus indicator when the user tabs through the buttons.

frontend/src/components/DefenceBox/DefenceMechanism.tsx
```diff
@@ -72,35 +79,49 @@ function DefenceMechanism({
 		>
 			<summary>
 				<span aria-hidden>{defenceDetail.name}</span>
```
gsproston-scottlogic, line 81: Interestingly, "prompt enclosure" isn't read out by the narrator when it's focused. I think it's because the checkbox is required for that. Just tried it out and one solution is to not hide the name when the defence is prompt enclosure.

```suggestion
				<span aria-hidden={defenceDetail.id !== DEFENCE_ID.PROMPT_ENCLOSURE}>
					{defenceDetail.name}
				</span>
```

gsproston-scottlogic (changes requested):
Really like this! Especially good work on the tests. Just a few notes from me. 

Regarding the concerns:

- Was unsure what to name this, went with prompt enclosure **- works for me** ✅ 
- I've got an generic info text "Alter the users input by appending instructional prompts and/or tags to the input." for Prompt Enclosure, then a shortened specific line for XML tagging/RSE. do we want to get rid of the generic one? **- I like it how it currently is, but not too fussed either way**
- Not sure about 'None' option - is it accessible etc? **- I'll leave this for Doro, but another solution could be to remove the "none" radio button, and have the checkbox slider like the other defences. I don't mind either way.**

frontend/src/components/ControlPanel/ControlPanel.tsx
```diff
@@ -28,21 +29,13 @@ function ControlPanel({
 }) {
 	function getDefencesConfigure() {
 		return defences.filter((defence) => {
-			return ![
-				DEFENCE_ID.PROMPT_EVALUATION_LLM,
-				DEFENCE_ID.QA_LLM,
-				DEFENCE_ID.SYSTEM_ROLE,
-			].some((id) => id === defence.id);
+			return !MODEL_DEFENCES.some((id) => id === defence.id);
 		});
```
chriswilty, line 33: Don't need curly braces when all an arrow function does is return something:

```
return defences.filter((defence) =>
  !MODEL_DEFENCES.some((id) => id === defence.id)
);
```

frontend/src/components/DefenceBox/DefenceBox.tsx
```diff
@@ -21,13 +22,25 @@ function DefenceBox({
...
+
 	return (
 		<div className="defence-box">
-			{defences.map((defence, index) => {
+			{getNotPromptEnclosureDefences().map((defence, index) => {
 				return (
 					<DefenceMechanism
 						key={index}
 						defenceDetail={defence}
+						promptEnclosureDefences={getPromptEnclosureDefences()}
```
chriswilty, line 43: DefenceMechanism was nice and generic previously; now it knows specifically about prompt enclosure defences.

I think this needs a cleaner solution, I'll have a think.

frontend/src/components/ThemedButtons/DefenceConfigurationRadioButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+		<div className="defence-radio-button" key={id}>
+			<input
+				type="radio"
+				id={id}
+				name={name}
+				value={id}
+				checked={checked}
+				onChange={onChange}
+			/>
+			<label htmlFor={id}>{name}</label>
```
chriswilty, line 24: Where possible, prefer enclosing inputs inside labels, rather than using ids. According to MDN, the label text may go after the input:

```
<label>
  <input ... />
  {name}
</label>
```

frontend/src/components/ThemedButtons/DefenceConfigurationRadioButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+	checked,
+	onChange,
+}: {
+	id: string;
+	name: string;
+	checked: boolean;
+	onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+}) {
+	return (
+		<div className="defence-radio-button" key={id}>
```
chriswilty, line 15: `key` is not needed here.

frontend/src/models/chat.ts
```diff
@@ -63,6 +63,7 @@ interface TransformedChatMessage {
 	preMessage: string;
 	message: string;
 	postMessage: string;
+	tranformationName: string;
```
chriswilty, line 66: Typo: `transformationName`

frontend/src/service/defenceService.ts
```diff
@@ -37,6 +37,7 @@ async function configureDefence(
 	config: DefenceConfigItem[],
 	level: number
 ): Promise<boolean> {
+	console.log('configuring', defenceId, config, level);
```
chriswilty, line 40: Just checking you still want this log statement?

chriswilty (changes requested):
A few concerns, but a great start.

backend/src/defence.ts
```diff
@@ -175,6 +196,35 @@ function transformXmlTagging(
...
 		message: escapeXml(message),
 		postMessage: closeTag,
+		tranformationName: 'XML Tagging',
+	};
+}
+
+function generateRandomString(string_length: number) {
+	let random_string = '';
+	for (let i = 0; i < string_length; i++) {
+		const random_ascii: number = Math.floor(Math.random() * 25 + 97);
```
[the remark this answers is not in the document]
heatherlogan-scottlogic: busted 🤐

frontend/src/Defences.ts
```diff
@@ -52,9 +52,26 @@ const DEFENCES_SHOWN_LEVEL3: Defence[] = [
...
+		[
+			makeDefenceConfigItem('PROMPT', 'prompt', 'text'),
+			makeDefenceConfigItem('SEQUENCE_LENGTH', 'sequence length', 'number'),
+		]
+	),
+
+	makeDefence(
+		DEFENCE_ID.PROMPT_ENCLOSURE,
+		'Prompt Enclosure',
+		"Alter the user's input by appending instructional prompts and/or tags to the input.",
```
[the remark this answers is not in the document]
heatherlogan-scottlogic: Good point although there is another Instruction defence we will add in which only adds a prompt to the start. What about "attaching"?

frontend/src/components/DefenceBox/DefenceBox.tsx
chriswilty: Added #745 for this.

chriswilty (approved):
👍 All good, added #745 to revisit the grouped defences.

PR #728, opened 2023-12-22
657 dont expose vitest globals

pmarsh-scottlogic:
## Description

Describe what your code changes do.

## Screenshots

## Notes

- Bullet point list with
- any notable code changes

## Concerns

- Bullet point list
- with any questions
- or concerns about this PR

## Checklist

Have you done the following?

- [ ] Linked the relevant Issue
- [ ] Added tests

### DIFF
--- a/backend/.eslintrc.cjs
+++ b/backend/.eslintrc.cjs
@@ -21,7 +21,7 @@ module.exports = {
 	parserOptions: {
 		project: ['./tsconfig.json', './test/tsconfig.json'],
 	},
-	plugins: ['@typescript-eslint'],
+	plugins: ['@typescript-eslint', 'jest'],
 	ignorePatterns: ['build', 'coverage', 'node_modules'],
 	rules: {
 		eqeqeq: 'error',
--- a/backend/test/integration/chatController.test.ts
+++ b/backend/test/integration/chatController.test.ts
@@ -25,7 +25,8 @@ declare module 'express-session' {
 }
 
 // mock the api call
-const mockCreateChatCompletion = jest.fn();
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const mockCreateChatCompletion = jest.fn<any>();
 jest.mock('openai', () => ({
 	OpenAI: jest.fn().mockImplementation(() => ({
 		chat: {
--- a/backend/test/integration/chatController.test.ts
+++ b/backend/test/integration/chatController.test.ts
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/unbound-method */
+import { expect, jest, test, describe } from '@jest/globals';
 import { Response } from 'express';
 
 import { handleChatToGPT } from '@src/controller/chatController';
--- a/backend/test/integration/chatController.test.ts
+++ b/backend/test/integration/chatController.test.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/unbound-method */
 import { expect, jest, test, describe } from '@jest/globals';
 import { Response } from 'express';
--- a/backend/test/integration/langchain.test.ts
+++ b/backend/test/integration/langchain.test.ts
@@ -25,17 +25,17 @@ import {
 } from '@jest/globals';
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
-const mockCall = jest.fn();
+const mockCall = jest.fn<any>();
 const mockRetrievalQAChain = {
 	call: mockCall,
 };
 const mockPromptEvalChain = {
 	call: mockCall,
 };
-const mockFromLLM = jest.fn();
-const mockFromTemplate = jest.fn();
+const mockFromLLM = jest.fn<any>();
+const mockFromTemplate = jest.fn<typeof PromptTemplate.fromTemplate>();
 const mockLoader = jest.fn();
-const mockSplitDocuments = jest.fn();
+const mockSplitDocuments = jest.fn<any>();
 const mockAsRetriever = jest.fn();
 
 // eslint-disable-next-line prefer-const
@@ -88,7 +88,6 @@ jest.mock('langchain/text_splitter', () => {
 		}),
 	};
 });
-
 // mock PromptTemplate.fromTemplate static method
 jest.mock('langchain/prompts');
 PromptTemplate.fromTemplate = mockFromTemplate;
@@ -113,7 +112,8 @@ RetrievalQAChain.fromLLM = mockFromLLM;
 
 jest.mock('@src/openai', () => {
 	// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-	const originalModule = jest.requireActual('@src/openai');
+	const originalModule =
+		jest.requireActual<typeof import('@src/openai')>('@src/openai');
 	// eslint-disable-next-line @typescript-eslint/no-unsafe-return
 	return {
 		...originalModule,
--- a/backend/test/integration/langchain.test.ts
+++ b/backend/test/integration/langchain.test.ts
@@ -15,6 +15,14 @@ import { RetrievalQAChain } from 'langchain/chains';
 import { PromptTemplate } from 'langchain/prompts';
 
 import { ChatOpenAI } from 'langchain/chat_models/openai';
+import {
+	afterEach,
+	beforeEach,
+	describe,
+	test,
+	jest,
+	expect,
+} from '@jest/globals';
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 const mockCall = jest.fn();
--- a/backend/test/integration/langchain.test.ts
+++ b/backend/test/integration/langchain.test.ts
@@ -1,4 +1,15 @@
-/* eslint-disable import/order */
+import {
+	afterEach,
+	beforeEach,
+	describe,
+	test,
+	jest,
+	expect,
+} from '@jest/globals';
+import { RetrievalQAChain } from 'langchain/chains';
+import { ChatOpenAI } from 'langchain/chat_models/openai';
+import { PromptTemplate } from 'langchain/prompts';
+
 import {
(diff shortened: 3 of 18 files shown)


backend/test/integration/chatController.test.ts
```diff
@@ -24,7 +24,8 @@ declare module 'express-session' {
 }
 
 // mock the api call
-const mockCreateChatCompletion = jest.fn();
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const mockCreateChatCompletion = jest.fn<any>();
```
chriswilty, line 28: Why do you need to use an any here, when there's a jest.fn() call below that doesn't specify a generic type?

backend/test/integration/langchain.test.ts
```diff
@@ -11,23 +22,19 @@ import {
...
-import { ChatOpenAI } from 'langchain/chat_models/openai';
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
-const mockCall = jest.fn();
+const mockCall = jest.fn<any>();
 const mockRetrievalQAChain = {
 	call: mockCall,
 };
 const mockPromptEvalChain = {
 	call: mockCall,
```
chriswilty, line 32: The problem seems to be that we are needlessly using the same mock function for two different calls. In fact, we are using it for three calls, as we also use it inline in the LLMChain constructor mock. Can we not declare these using inline jest.fn()?

```
const mockPromptEvalChain = {
  call: jest.fn(),
};

expect(mockPromptEvalChain.call).toHaveBeenCalled();
```

backend/test/integration/langchain.test.ts
```diff
@@ -11,23 +22,19 @@ import {
...
 };
-const mockFromLLM = jest.fn();
-const mockFromTemplate = jest.fn();
+const mockFromLLM = jest.fn<any>();
+const mockFromTemplate = jest.fn<typeof PromptTemplate.fromTemplate>();
```
chriswilty, line 35: Is this correct? The type args of jest.fn, where given, appear to be the return value and then the args of the function. Therefore I would expect this either to be an untyped `jest.fn()`, or

```
jest.fn<
  ReturnType<typeof PromptTemplate.fromTemplate>,
  Parameters<typeof PromptTemplate.fromTemplate>
>();
```

Or are the types provided by @jest/globals different to those provided by @types/jest?

chriswilty (commented):
Unsure about a couple of things. In general, it seems this PR is tackling more than what is described in the issue. If we are deliberately _not_ providing jest functions in global scope, then we don't really need any new linter config, because the tests wouldn't compile if we tried to use the jest functions without importing them, right?

backend/test/integration/chatController.test.ts
pmarsh-scottlogic: If we do not set a type then each time we want to mock its behaviour we get an error:

image

image

backend/test/integration/langchain.test.ts
pmarsh-scottlogic: I tried your suggestion and it gets this error:

image

If we look at [the docs](image), it looks as though the type arg is supposed to be the type of the method we are mocking, which is how I did the other non-generic type args.

Relevant examples from the doc:
image
image
image

backend/test/integration/langchain.test.ts
pmarsh-scottlogic: My latest commit has resolved this :)

pmarsh-scottlogic:
> Unsure about a couple of things. In general, it seems this PR is tackling more than what is described in the issue. If we are deliberately _not_ providing jest functions in global scope, then we don't really need any new linter config, because the tests wouldn't compile if we tried to use the jest functions without importing them, right?

You're right, the linter plugin is beyond scope and the program won't compile if we are missing the imports. The purpose of the linter plugin is so we can auto import stuff from the right place - i.e. if we haven't imported the jest object already, then it will get highlighted, and then we can right click -> import from jest/globals, to save us from having to type it manually. Without the plugin, eslint doesn't know where to find the right import. It's just a quality of life thing - the type issues all come out of typescript, rather than the linter.

backend/test/integration/langchain.test.ts
chriswilty: Ok, so `@types/jest` is in fact wrong! Good to know.

chriswilty:
> > Unsure about a couple of things. In general, it seems this PR is tackling more than what is described in the issue. If we are deliberately _not_ providing jest functions in global scope, then we don't really need any new linter config, because the tests wouldn't compile if we tried to use the jest functions without importing them, right?
> 
> You're right, the linter plugin is beyond scope and the program won't compile if we are missing the imports. The purpose of the linter plugin is so we can auto import stuff from the right place - i.e. if we haven't imported the jest object already, then it will get highlighted, and then we can right click -> import from jest/globals, to save us from having to type it manually. Without the plugin, eslint doesn't know where to find the right import. It's just a quality of life thing - the type issues all come out of typescript, rather than the linter.

Hmm, I guess that's a weakness of VSCode then, as I get import suggestions in WebStorm 🙄

backend/test/integration/chatController.test.ts
chriswilty: Using `any` is not helping either, cos it tells us nothing about how the mock is expected to be used (and forces us to disable a perfectly good linter rule). What values are we expecting to be returned from createChatCompletion? Can we not provide a union type to the function for that?

pmarsh-scottlogic: indeed. I've removed all the any types now!

backend/test/integration/chatController.test.ts
```diff
@@ -42,50 +47,50 @@ function responseMock() {
...
+			{
+				message: {
+					tool_calls: [
+						{
+							type: 'function',
+							id: 'sendEmail',
+							function: {
+								name: 'sendEmail',
+								arguments: JSON.stringify({
+									...testSentEmail,
```
pmarsh-scottlogic, line 81: I've had to move `chatSendEmailResponseAssistant`, `chatResponseAssistant` and `chatSendEmailResponseAssistant` so that I can use them to define the type of `mockCreateChatCompletion` (line 26). The other option would be to keep them where they are and write a manual type definition, but then you run the risk of them coming out of sync...
Neither option seems particularly pretty.

backend/test/integration/langchain.test.ts
```diff
@@ -209,11 +210,9 @@ describe('langchain integration tests ', () => {
 	});
 
 	test('GIVEN the prompt evaluation model is initialised WHEN it is asked to evaluate an input AND it does not respond in the correct format THEN it returns a final decision of false', async () => {
-		mockFromLLM.mockImplementation(() => mockPromptEvalChain);
```
pmarsh-scottlogic, line 212: So, fromLLM is basically a fancy static method within the `RetrievalQAChain` class that constructs an instance of `RetrievalQAChain`. Therefore every time we mock it to return the `mockPromptEvalChain`, it was wrong.

Further, since it should actually return `mockRetrievalQAChain` for every test, we don't need to mock differently it per test, so I've moved the mockImplementation to the beforeEach

backend/test/integration/openai.test.ts
```diff
@@ -13,9 +13,10 @@ import { EmailInfo } from '@src/models/email';
 import { chatGptSendMessage } from '@src/openai';
 import { systemRoleDefault } from '@src/promptTemplates';
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const mockCreateChatCompletion = jest.fn<any>();
-// Mock the OpenAIApi class
+const mockCreateChatCompletion =
+	jest.fn<() => Promise<ReturnType<typeof chatResponseAssistant>>>();
```
chriswilty, line 17: Typescripting like a pro 🫶

backend/.eslintrc.cjs
```diff
@@ -67,4 +67,14 @@ module.exports = {
 		'@typescript-eslint/consistent-type-definitions': 0, // disable rule. Eventually use below rule to enforce type over interface
 		// '@typescript-eslint/consistent-type-definitions': ['error', 'type'],
 	},
+	overrides: [
+		{
+			files: ['*.test.ts'],
+			plugins: ['jest'],
+			rules: {
+				'@typescript-eslint/unbound-method': 'off',
+				'jest/unbound-method': 'error',
```
chriswilty, line 76: Nice find 👍

chriswilty (approved):
Nice work 👌

backend/test/integration/langchain.test.ts
```diff
@@ -95,18 +99,16 @@ jest.mock('langchain/chains', () => {
...
-			return {
-				call: mockCall,
-			};
+			return mockPromptEvalChain;
 		}),
 	};
 });
-RetrievalQAChain.fromLLM = mockFromLLM;
+RetrievalQAChain.fromLLM =
+	mockFromLLM as unknown as typeof RetrievalQAChain.fromLLM;
```
gsproston-scottlogic, line 107: I suppose we can't avoid the `as unknown` ? I vaguely remember some discussions about it but can't remember where we landed.

gsproston-scottlogic (approved):
All looks good! I have a question but nothing stopping me from approving.

backend/test/integration/langchain.test.ts
pmarsh-scottlogic: I think a cleverer person could avoid it.

As it stands though
- We want to mock this static method `RetrievalQAChain.fromLLM`
- `RetrievalQAChain.fromLLM` has its set type defined by the library (a function type which takes a bunch of complicated inputs and outputs an instance of `RetrievalQAChain`)
- Our mock will be a function that takes no inputs and outputs a particular mocked version of `RetrievalQAChain`. I want to keep it as simple as possible, and therefore it has a different type to 
- The one way that I've found to mock a static method so far that works is to import the object `RetrievalQAChain` and reassign its value explicitly `RetrievalQAChain.fromLLM = ...` [pretty sure there are other ways to do this]. But because the types differ, this requires casting, and because of some typescript or eslint quirk, it requires `as unknown as`

gsproston-scottlogic: Good to know, thanks!

PR #727, opened 2023-12-22
699 Update responsive design up to 200% zoom

Kenny4297:
## Description

I have implemented media queries across all CSS files pertaining to the Header. These modifications ensure proper display and functionality of the header at up to 200% zoom.

## Screenshots

Here's how the header looks at 200% zoom:
200Zoom

## Notes

While this project's structure—with multiple components and corresponding CSS files—works well with JavaScript logic implementation, it presents challenges in styling, particularly for the header.

## Concerns

In my personal opinion, I would be worried about the scalability of this project when it comes to responsive design. Although it is possible to make this application responsive, the fact that multiple CSS files are used (and scattered) throughout the application with regards to one section of the application (the header, for example) will make responsive design difficult.

Thanks for letting me contribute!!

PR #729, opened 2023-12-23
Responsive zoom issue#699

Kenny4297:
## Description

Here is the most recent update for my previous issue. The only change from my last PR is running the formatter on the front end.

I have implemented media queries across all CSS files pertaining to the Header. These modifications ensure proper display and functionality of the header at up to 200% zoom.

## Screenshots

Here's how the header looks at 200% zoom:
200Zoom

## Notes

While this project's structure—with multiple components and corresponding CSS files—works well with JavaScript logic implementation, it presents challenges in styling, particularly for the header.

## Concerns

In my personal opinion, I would be worried about the scalability of this project when it comes to responsive design. Although it is possible to make this application responsive, the fact that multiple CSS files are used (and scattered) throughout the application with regards to one section of the application (the header, for example) will make responsive design difficult.

Thanks for letting me contribute!!

### DIFF
--- a/frontend/src/components/LevelSelectionBox/LevelSelectionBox.css
+++ b/frontend/src/components/LevelSelectionBox/LevelSelectionBox.css
@@ -12,31 +12,31 @@
 
 @media (max-width: 1230px) {
 	.level-selection-box button {
-		font-size: .8rem;
+		font-size: 0.8rem;
 		padding: 1rem;
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pmarsh-scottlogic:
Hi @Kenny4297  I believe you don't need to open this new PR. I can see the new changes under the original PR #727 ! It would be preferable just to keep that one so that all the history gets captured in one place :)

PR #733, opened 2024-01-03
598 add tooltips to disabled buttons

gsproston-scottlogic:
## Description

Adding tooltips (titles) to various buttons when they're disabled.

## Screenshots

Level buttons (hovering over level 2):

image

Send button: 

image

Model change:

image

## Notes

- The level select buttons aren't actually ThemedButtons anymore, but this was easy enough to add in.

## Concerns

- Before, the `LevelButton` component didn't have a `title`, but did have an `aria-label`. Now, we have an `aria-label` that is not the same as the `title` when the button is disabled. Is this acceptable?
- Before, the `LoadingButton` component didn't have either a `title` or an `aria-label`. Should add an `aria-label` as well when the button is disabled?
- There can also be disabled buttons in the document viewer. For example, when the first document is displayed, the previous document button is disabled. I didn't add tooltips for these as I thought it was clear why they would be disabled, but happy to add them in if necessary. 

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/frontend/src/components/ChatBox/ChatBox.tsx
+++ b/frontend/src/components/ChatBox/ChatBox.tsx
@@ -222,6 +222,7 @@ function ChatBox({
 						<LoadingButton
 							onClick={() => void sendChatMessage()}
 							isLoading={isSendingMessage}
+							loadingTooltip="Sending message..."
 						>
 							send
 						</LoadingButton>
--- a/frontend/src/components/LevelSelectionBox/LevelSelectionBox.tsx
+++ b/frontend/src/components/LevelSelectionBox/LevelSelectionBox.tsx
@@ -25,20 +25,25 @@ function LevelSelectionBox({
 			setCurrentLevel(newLevel);
 		}
 	}
+
 	return (
 		<div className="level-selection-box">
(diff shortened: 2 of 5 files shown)


dhinrichs-scottlogic:
Alright, we had some discussions and here's what we came up with:
- NVDA reads out both the aria-label and the title, so for the level buttons that is: "Level 3 button unavailable complete level 2 to unlock" and for the send button that is: "send button unavailable sending message".
- Narrator only reads out the aria-label, but not the title. ("Level 3 button disabled" and "send button disabled")
- We considered adding the text "complete level 2 to unlock" to the aria-label, and decided against it, since this would lead to NVDA repeating that text twice (from the aria-label and then the title). We think that the title is important for visual users, since the visual design of the disabled button might not be clear enough for some users (especially since the "Level" text and the level number are separate). We hope that "Level 3 button disabled" is a clear enough message for Narrator users, and since NVDA is more commonly used than Narrator, we think this is the best accessible option we have at the moment.

dhinrichs-scottlogic (approved):
Looks good!

heatherlogan-scottlogic (approved):
Nice 🤓

PR #737, opened 2024-01-03
706 refactor updateconfigproperty to remove mutation

heatherlogan-scottlogic:
## Description

Smol refactor based on Peter's nice suggestions on ticket 🤓
Didn't think it was necessary to check oldValue=newValue in controller as the slider's don't send requests on unchanged values  

## Screenshots

## Notes

- Bullet point list with
- any notable code changes

## Concerns

- Bullet point list
- with any questions
- or concerns about this PR

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/backend/src/controller/modelController.ts
+++ b/backend/src/controller/modelController.ts
@@ -13,8 +13,7 @@ function updateConfigProperty(
 	max: number
 ): ChatModelConfiguration | null {
 	if (value >= 0 && value <= max) {
-		config[configId] = value;
-		return config;
+		return { ...config, [configId]: value };
 	}
 	return null;
 }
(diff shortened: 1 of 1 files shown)


backend/src/controller/modelController.ts
```diff
@@ -37,32 +36,23 @@ function handleConfigureModel(req: OpenAiConfigureModelRequest, res: Response) {
 	const configId = req.body.configId as MODEL_CONFIG | undefined;
 	const value = req.body.value;
 
-	let updated = null;
-
-	if (configId && value && value >= 0) {
+	if (configId && value !== undefined) {
```
gsproston-scottlogic, line 39: Good catch, just checked and this does indeed fix a bug in dev where we couldn't set 0 values for these.

backend/src/controller/modelController.ts
```diff
@@ -37,32 +36,23 @@ function handleConfigureModel(req: OpenAiConfigureModelRequest, res: Response) {
...
-				break;
-			case MODEL_CONFIG.PRESENCE_PENALTY:
-				updated = updateConfigProperty(lastConfig, configId, value, 2);
-				break;
-			default:
-				res.status(400).send();
-		}
-		if (updated) {
-			req.session.chatModel.configuration = updated;
+		const maxValue = configId === MODEL_CONFIG.TOP_P ? 1 : 2;
```
gsproston-scottlogic, line 41: I'd personally be tempted to move this above the `if` statement on line 39, and then add a range check to that `if` statement. 
Then, you don't even need the `updateConfigProperty` method anymore, just do 

```typescript
function handleConfigureModel(req: OpenAiConfigureModelRequest, res: Response) {
	const configId = req.body.configId as MODEL_CONFIG | undefined;
	const value = req.body.value;
	const maxValue = configId === MODEL_CONFIG.TOP_P ? 1 : 2;

	if (configId && value !== undefined && value >= 0 && value <= maxValue) {
		req.session.chatModel.configuration[configId] = value;
		res.status(200).send();
	} else {
		res.status(400).send();
	}
}
```

gsproston-scottlogic (changes requested):
Nicely trimmed down. I actually just thought of a way we could simplify it even further.

chriswilty (commented):
What George said! Otherwise good to go.

backend/src/controller/modelController.ts
heatherlogan-scottlogic: ah yeah much nicer!

gsproston-scottlogic (approved):
Brill!

chriswilty (approved):
Beautiful!

Issue #723, opened 2023-12-22
Redesign the left panel


gsproston-scottlogic:
# Feature Request

## Description
[update: some of the configs have been redesigned for accessibility, here's the design updates that could be made to update the rest] 

- Make scrollbar invisible so that it doesn't shift the content as it appears
- Ensure <textarea>s and <input type=number> have sufficient colour contrast, even when readOnly
- <s>Give `<summary>` buttons the info icon (make sure it's aria-hidden) and apply the same button across all configurations</s> instead: replace the "i" icon button in the model configuration with the "Detail" button that the defence configurations have
- Give the model selection its own visibly distinct area (e.g. a fieldset like the others?)
- Adjust the model configurations that have sliders (e.g. Model Temperature, Top P ...) so that they look like the other configurations (e.g. Prompt Evaluation LLM). it's already a <fieldset>, just needs the same border, headings and buttons. 

## Screenshot of suggested changes:
![image](https://github.com/ScottLogic/prompt-injection/assets/125262707/81508184-7c31-48ba-bb5f-c552e6777ed7)

## Acceptance criteria
- Given a user is in sandbox mode

**Scroll bar:**
- When they expand multiple configurations so that scrolling is necessary
- Then  the sidepanel width does not shrink to make space for the scrollbar (either because the scroll is invisible, or because the space is adjusted to take the scrollbar into account)

**Colour Contrast:**
- When the input fields of the configurations are in read only
- Then the colour contrast between the background and the text of the input element is at least 4.5:1

**Model Selection area**
- When the user interacts with the Model Selection are
- Then the area is visually programmatically grouped together (you can see the controls belong together and a screen reader will tell you the controls are grouped by announcing something like "Model Selection, grouping")

**Model Configurations (e.g. Model Temperature, Top P, all of those with slider controls)**
- When the user interacts with the Model Configurations 
- Then the area is visually programmatically grouped together (you can see the controls belong together and a screen reader will tell you the controls are grouped by announcing something like "Model Temperature, grouping") 
AND
- The button to expand the section for more information (currently an "i" icon) looks identical to the "Details" button in the Defence Configurations

chriswilty:
@dhinrichs-scottlogic to update with latest state and ideas for improvement / UX input.

chriswilty:
Like these 👍 

Do we want to change slider colour to the theme teal at the same time? It's odd that they're the only green controls.

gsproston-scottlogic:
> Like these 👍
[...]

PR #748, opened 2024-01-08
725 Toggle button and defence config restyle

dhinrichs-scottlogic:
## Description
Redesigned the toggle buttons and Configurations on the left side to ensure Screen Reader support (NVDA for Firefox & Chrome, Narrator for Edge).

## Screenshots
image

## Notes
- Defence configurations grouped into fieldsets, which allows for multiple input controls to be clearly linked for Assistive Technology. 
- Toggle buttons reworked giving them a visual label (on/off)
- Summary button (the button that expands to see more info) reworked to look like a button

## Concerns 
(would prefer to address in separate tickets to keep this PR small, but noticed as part of this)
- The Model Temperature configurations could also be changed to use this new design for consistently
- The tabbing order is slightly controversial: The on/off button comes before the details button, it might be expected to be the other way round since we read left to right, although you could also argue that the on/off is the main control (and slightly higher up than the details button) and that the details button secondary. I tried to switch them around, but this creates issues for the layout when the details section is expanded. Could make this order more clear with nudging the on/off toggle a bit further up. Personally I think it's not perfect, but good enough. Please test this out by tabbing through the controls and let me know what you think.
image
- <s>The text/number inputs of the defences are not available to screen reader when the defence is inactive. </s> really easy fix, so not a concern anymore

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/frontend/src/components/DefenceBox/DefenceMechanism.css
+++ b/frontend/src/components/DefenceBox/DefenceMechanism.css
@@ -18,9 +18,8 @@
 	border-radius: 0.25rem;
 }
 
-/* style for when details element is expanded */
-.defence-mechanism-fieldset details[open] > summary {
-	border-style: inset;
+.defence-mechanism > summary {
+	font-size: 1rem;
 }
 
 /* remove the default triangle on the summary element */
@@ -32,6 +31,11 @@
 	display: none;
 }
 
+/* style for when details element is expanded */
+.defence-mechanism-fieldset details[open] > summary {
+	border-style: inset;
+}
+
 /* the toggle and label combined */
 .defence-mechanism-form {
 	position: absolute;
@@ -40,14 +44,6 @@
 
 .defence-mechanism {
 	cursor: default;
-}
-
-.defence-mechanism > summary {
-	font-size: 1rem;
-}
-
-.defence-mechanism .info-box {
-	font-size: 0.875rem;
 }
 
 .defence-mechanism .info-box :first-child {
@@ -98,31 +94,18 @@
 	text-align: right;
 }
 
-/* the  label hovered or focused */
-.toggles [type='checkbox']:focus + label,
-.toggles [type='checkbox']:hover + label {
-	color: var(--main-text-accent-colour);
-}
-
-.toggles [type='checkbox']:focus + label::before,
-.toggles [type='checkbox']:hover + label::before {
-	box-shadow: 0 0 0.3em var(--main-text-colour);
-}
-
-.toggles [type='checkbox']:focus + label::after,
-.toggles [type='checkbox']:hover + label::after {
-	background-image: url("data:image/svg+xml,%3Csvg viewBox='0 0 100 100' xmlns='http://www.w3.org/2000/svg'%3E%3Ccircle cx='50' cy='50' r='50' fill='rgba(0,0,0,.25)'/%3E%3C/svg%3E");
-	background-position: center center;
-	background-size: 30%;
-	background-repeat: no-repeat;
-}
-
 .toggles [type='checkbox'] + label::before,
 .toggles [type='checkbox'] + label::after {
 	content: '';
 	position: absolute;
 	height: 0.9em;
 	transition: all 0.25s ease;
+}
+
+/* the  label hovered or focused */
+.toggles [type='checkbox']:focus + label,
+.toggles [type='checkbox']:hover + label {
+	color: var(--main-text-accent-colour);
 }
 
 /* background oval before it's checked */
@@ -144,6 +127,19 @@
 	border-radius: 50%;
 	background: var(--main-text-colour);
 	background-position: center center;
+}
+
+.toggles [type='checkbox']:focus + label::before,
+.toggles [type='checkbox']:hover + label::before {
+	box-shadow: 0 0 0.3em var(--main-text-colour);
+}
+
+.toggles [type='checkbox']:focus + label::after,
+.toggles [type='checkbox']:hover + label::after {
+	background-image: url("data:image/svg+xml,%3Csvg viewBox='0 0 100 100' xmlns='http://www.w3.org/2000/svg'%3E%3Ccircle cx='50' cy='50' r='50' fill='rgba(0,0,0,.25)'/%3E%3C/svg%3E");
+	background-position: center center;
+	background-size: 30%;
+	background-repeat: no-repeat;
 }
 
 /* round toggle slider sliding to the right */
--- a/frontend/src/components/DefenceBox/DefenceMechanism.css
+++ b/frontend/src/components/DefenceBox/DefenceMechanism.css
@@ -41,10 +41,6 @@
 	cursor: default;
 }
 
-/* .defence-mechanism > * {
-	padding: 0.75rem;
-} */
-
 .defence-mechanism > summary {
 	font-size: 1rem;
 }
@@ -53,9 +49,9 @@
 	font-size: 0.875rem;
 }
 
-/* .defence-mechanism .info-box :first-child {
+.defence-mechanism .info-box :first-child {
 	margin-top: 0;
-} */
+}
 
 /* the expanded description and further controls */
 .defence-mechanism .info-box {
@@ -160,18 +156,6 @@
 .toggles [type='checkbox']:checked + label::before {
 	background-color: var(--main-toggle-on-border-colour);
 	border-color: var(--main-toggle-on-border-colour);
-}
-
-/* Windows High Contrast Mode Support */
-@media screen and (-ms-high-contrast: active) {
-	.toggles [type='checkbox']:focus + label::before,
-	.toggles [type='checkbox']:hover + label::before {
-		outline: 1px dotted windowText;
-		outline-offset: 0.25em;
-	}
-	.toggles [type='checkbox'] + label::after {
-		background-color: windowText;
-	}
 }
 
 /* Reduced motion */
--- a/frontend/src/components/DefenceBox/DefenceMechanism.css
+++ b/frontend/src/components/DefenceBox/DefenceMechanism.css
@@ -1,92 +1,66 @@
+.defence-mechanism-fieldset {
+	border: 0.25rem solid var(--control-config-border);
+	border-radius: 0.5rem;
+	margin: 0.1em;
+	min-height: fit-content;
+}
+
+/* the name of the fieldset */
+.defence-mechanism-legend {
+	padding: 0.3em;
+}
+
+/* the more button to expand the details */
+.defence-mechanism-summary {
+	width: fit-content;
+	border: 0.125rem outset var(--control-config-border);
+	border-radius: 0.25rem;
+	padding: 0.5rem;
+}
+
+/* style for when details element is expanded */
+.defence-mechanism-fieldset details[open] > summary {
+	border-style: inset;
+}
+
+/* remove the default triangle on the summary element*/
+.defence-mechanism-fieldset details > summary {
+	list-style: none;
+}
+.defence-mechanism-fieldset details > summary::-webkit-details-marker {
+	display: none;
+}
+
+/* the toggle and label combined */
+.defence-mechanism-form {
+	position: absolute;
+	right: 1em;
+}
+
 .defence-mechanism {
 	cursor: default;
 }
 
-.defence-mechanism > * {
+/* .defence-mechanism > * {
 	padding: 0.75rem;
-}
+} */
 
 .defence-mechanism > summary {
-	position: relative;
 	font-size: 1rem;
-}
-
-/*
- * Modified from https://www.w3schools.com/howto/howto_css_switch.asp
-*/
-
-/* The switch - the box around the slider */
-.defence-mechanism label.switch {
-	position: absolute;
-	top: 0.75rem;
-	right: 0.75rem;
-	width: 2rem;
-	height: 1rem;
-}
-
-.defence-mechanism label.switch:focus-within {
-	border-radius: 0.5rem;
-	outline-style: auto;
-}
-
-/* Hide default HTML checkbox */
-.defence-mechanism label.switch > input {
-	width: 0;
-	height: 0;
-	opacity: 0;
-}
-
-/* The slider */
-.defence-mechanism label.switch > .slider {
-	--slider-border-width: 0.125rem;
-
-	position: absolute;
-	inset: 0;
-	border: var(--slider-border-width) solid var(--main-toggle-off-border-colour);
-	background-color: var(--main-toggle-off-colour);
-	cursor: pointer;
-	transition: 0.4s;
-}
-
-.defence-mechanism label.switch > .slider::before {
-	content: '';
-	position: absolute;
-	top: calc(-1 * var(--slider-border-width));
-	left: calc(-1 * var(--slider-border-width));
-	width: 1rem;
-	height: 1rem;
-	background-color: var(--main-toggle-ball-colour);
-	transition: 0.4s;
-}
-
-.defence-mechanism label.switch > input:focus + .slider {
-	box-shadow: 0 0 0.0625rem var(--main-toggle-on-colour);
-}
-
-.defence-mechanism label.switch > input:checked + .slider {
-	border-color: var(--main-toggle-on-border-colour);
-	background-color: var(--main-toggle-on-colour);
-}
-
-.defence-mechanism label.switch > input:checked + .slider::before {
-	transform: translateX(1rem);
-}
-
-/* Rounded sliders */
-.defence-mechanism label.switch > .slider.round {
-	border-radius: 0.5rem;
-}
-
-.defence-mechanism label.switch > .slider.round::before {
-	border-radius: 50%;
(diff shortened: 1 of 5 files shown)


gsproston-scottlogic (approved):
All works for me 👍 
Regarding the concerns:

- The Model Temperature configurations could also be changed to use this new design for consistently **- fair, could be done in** #723 
- The tabbing order is slightly controversial: The on/off button comes before the details button, it might be expected to be the other way round since we read left to right, although you could also argue that the on/off is the main control (and slightly higher up than the details button) and that the details button secondary. I tried to switch them around, but this creates issues for the layout when the details section is expanded. Could make this order more clear with nudging the on/off toggle a bit further up. Personally I think it's not perfect, but good enough. Please test this out by tabbing through the controls and let me know what you think. **- I get that it might not be what's expected, but I think it's fine.**
- The text/number inputs of the defences are not available to screen readera when the defence is inactive. I created [a ticket](https://github.com/ScottLogic/prompt-injection/issues/750) to address this. **- interestingly, the only one I can't focus on is the character limit configuration, maybe because it's a number input?**

frontend/src/components/ThemedInput/ThemedNumberInput.tsx
```diff
@@ -43,7 +43,7 @@ function ThemedNumberInput({
 			className={inputClass}
 			type="number"
 			value={content}
-			disabled={disabled}
+			readOnly={disabled}
```
pmarsh-scottlogic, line 46: interesting! How does readOnly compare from aria-disabled? I think we're using aria-disabled elsewhere to indicate that something is disabled, but still tabbable.

frontend/src/components/DefenceBox/DefenceMechanism.css
```diff
@@ -1,106 +1,164 @@
...
 }
 
-.defence-mechanism label.switch > input:focus + .slider {
-	box-shadow: 0 0 0.0625rem var(--main-toggle-on-colour);
+/* the  label */
+.toggles [type='checkbox'] + label {
+	position: relative;
+	display: block;
+	max-width: fit-content;
+	padding-right: 3em;
```
pmarsh-scottlogic, line 93: Lots of `em` units showing up. I wonder if it's worth having them be rem to be more consistent with our other css? I've not looked into if it's important that they're em yet... If we want to change to rem, then we might want to check that the values convert to nice numbers of pixels (since base font is 16px)

frontend/src/components/DefenceBox/DefenceMechanism.tsx
```diff
@@ -69,73 +69,77 @@ function DefenceMechanism({
...
-			<summary>
-				<span aria-hidden={defenceDetail.id !== DEFENCE_ID.PROMPT_ENCLOSURE}>
-					{defenceDetail.name}
-				</span>
-				{defenceDetail.id !== DEFENCE_ID.PROMPT_ENCLOSURE && (
-					<label className="switch">
+		<fieldset className="defence-mechanism-fieldset">
+			<legend className="defence-mechanism-legend">{defenceDetail.name}</legend>
+			{defenceDetail.id !== DEFENCE_ID.PROMPT_ENCLOSURE && (
+				<form className="defence-mechanism-form">
```
pmarsh-scottlogic, line 75: I'm curious about the rationale of wrapping the toggles `<div>` within a form.

frontend/src/components/DefenceBox/DefenceMechanism.css
```diff
@@ -1,106 +1,164 @@
-.defence-mechanism {
-	cursor: default;
+.defence-mechanism-fieldset {
+	min-height: fit-content;
+	margin: 0.1em;
```
pmarsh-scottlogic, line 3: I feel like this margin is so small that it may as well not be there?

frontend/src/components/DefenceBox/DefenceMechanism.css
```diff
@@ -1,106 +1,164 @@
...
+	border: 0.25rem solid var(--control-config-border);
```
pmarsh-scottlogic, line 4: Great idea to add the border! Makes it clear where one config ends and the next begins :)

```suggestion
	border: 0.15rem solid var(--control-config-border);
```
I'd prefer a slightly less chonky border, but that's a really minor aesthetic choice, so feel free to ignore

frontend/src/components/DefenceBox/DefenceMechanism.css
pmarsh-scottlogic: In fact, I noticed a tiny tiny gap between the edge of the border and the edge of the defence-configuration box, which I'd prefer to not be there. Or if we want a gap, then I reckon it should be a bit bigger so that it looks more intentional.

image

So based on that, I'd actually prefer to have `margin: 0` to overwrite these tiny margins being added to the field set by 'user agent stylesheet' (what is that?).

image

image

frontend/src/components/DefenceBox/DefenceMechanism.css
```diff
@@ -1,106 +1,164 @@
...
+	flex-direction: column;
+	gap: 0.5rem;
+	padding: 0.75rem 0;
+}
+
+.defence-mechanism .prompt-enclosure-configuration-area {
+	padding: 0;
+}
+
+/* adapted from: https://adrianroselli.com/2019/03/under-engineered-toggles.html */
```
pmarsh-scottlogic, line 75: good thinking to include this. I'll take some time to read it fully!

frontend/src/components/DefenceBox/DefenceMechanism.css
```diff
@@ -1,106 +1,164 @@
...
-/* The switch - the box around the slider */
-.defence-mechanism label.switch {
+.defence-mechanism-fieldset details > summary::-webkit-details-marker {
+	display: none;
+}
+
+/* style for when details element is expanded */
+.defence-mechanism-fieldset details[open] > summary {
+	border-style: inset;
+}
```
pmarsh-scottlogic, line 37: this is clever. I but despite the comment, it's a bit hard to figure out what it's doing without inspecting in the browser...

Again, just an aesthetic point, but I'd prefer a flat button (to be consistent with rest of app), and then to indicate open or closed I'd put the little triangle (that used to be on the `<summary>`) on the button

frontend/src/components/DefenceBox/DefenceMechanism.css
```diff
@@ -1,106 +1,164 @@
...
-	opacity: 0;
+.defence-mechanism .info-box :first-child {
+	margin-top: 0;
 }
 
-/* The slider */
-.defence-mechanism label.switch > .slider {
-	--slider-border-width: 0.125rem;
+/* the expanded description and further controls */
+.defence-mechanism .info-box {
```
pmarsh-scottlogic, line 54: another case of removing a comment and improving the classname. Maybe `expanded-box` or `expanded-description-and-config-value`. I'd have to give it some more thought.

pmarsh-scottlogic (changes requested):
Love how the screenreader is behaving now! It feels super nice and clear what's going on. Didn't try with any other assistive technologies.

Re: tab ordering - I think it feels relatively natural, so I wouldn't worry about it.

A few requests from me. Notably there's a few aesthetic improvements that I'd like but most of them are subjective, and can be debated. Also a few instances of comments where I think naming could be clearer. I concede that I have a pretty strong view on this though, stemming from [this video](https://www.youtube.com/watch?v=Bf7vDBBOBUA).

## Focus problems
The focus indicators seem to be getting cut off at the edges for the main config `<summary>` elements.
image
image
image

not quite sure which bit of css is causing this.

frontend/src/components/DefenceBox/DefenceMechanism.css
dhinrichs-scottlogic: it's the same className that it currently is on dev, so didn't want to change that around unnecessarily, especially since we don't really have consistent naming convention throughout

frontend/src/components/DefenceBox/DefenceMechanism.css
dhinrichs-scottlogic: personally I think this is clearer than the triangle, especially since the triangle is already on the parent component, and the difference between expanded and a collapsed triangle is only a few degrees of rotation which is not that easy to see.
image

frontend/src/components/DefenceBox/DefenceMechanism.css
dhinrichs-scottlogic: In my last PR Chris made the point that it would be better to try to use whole-pixel equivalents of rems:
image
I tried a skinnier border but thought this looked better chunky.

frontend/src/components/DefenceBox/DefenceMechanism.css
dhinrichs-scottlogic: yep, agreed, I'll turn this to margin: 0;

frontend/src/components/DefenceBox/DefenceMechanism.css
```diff
@@ -1,106 +1,164 @@
...
+}
+
+/* the name of the fieldset */
+.defence-mechanism-legend {
+	padding: 0.3em;
 }
 
-.defence-mechanism > * {
-	padding: 0.75rem;
+/* the more button to expand the details */
```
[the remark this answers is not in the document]
dhinrichs-scottlogic: I'm happy to remove the comments for the different classes, I thought it would help as folks review and then someone would tell me to remove the comments 😄

frontend/src/components/DefenceBox/DefenceMechanism.css
```diff
@@ -1,106 +1,164 @@
...
+.defence-mechanism-fieldset {
+	min-height: fit-content;
+	margin: 0.1em;
+	border: 0.25rem solid var(--control-config-border);
+	border-radius: 0.5rem;
+}
+
+/* the name of the fieldset */
+.defence-mechanism-legend {
+	padding: 0.3em;
```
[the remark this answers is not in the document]
dhinrichs-scottlogic: the article where I found the accessible toggle button says this about ems:
"All sizing is done in ems so these will scale with the surrounding text."

frontend/src/components/DefenceBox/DefenceMechanism.css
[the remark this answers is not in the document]
pmarsh-scottlogic: @dhinrichs-scottlogic that's really thoughtful! Was pretty helpful actually for review :)

frontend/src/components/DefenceBox/DefenceMechanism.css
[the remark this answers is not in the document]
dhinrichs-scottlogic: unless we have a separate file for the toggles I'd say no, I'd find it confusing to have two separate css files for the same tsx file

frontend/src/components/DefenceBox/DefenceMechanism.tsx
```diff
@@ -69,73 +69,77 @@ function DefenceMechanism({
...
+					</div>
+				</form>
+			)}
+			<details
+				className="defence-mechanism"
+				onToggle={() => {
+					// re-render the configuration component when detail is toggled
+					// this is to resize the textarea when detail is expanded
+					setConfigKey(configKey + 1);
+				}}
```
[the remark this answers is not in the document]
dhinrichs-scottlogic: I think that was already there

frontend/src/components/DefenceBox/DefenceMechanism.css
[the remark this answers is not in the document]
pmarsh-scottlogic: Curious. I guess it makes sense that you want the toggle to match the size of the text that it is near.

The difference I can find is that `1 rem` is a size relative to the root font size and `em` is relative to the the font size of the element that you're styling (whether set explicitly or inherited from an ancestor). So in this case if we changed the size of the legend text... it will scale the padding accordingly. Which is great!

Consider my first comment nullified :)

frontend/src/components/DefenceBox/DefenceMechanism.css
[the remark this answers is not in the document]
pmarsh-scottlogic: that's fair. In that case I would advocate for a separate file for the toggle :)

frontend/src/components/ThemedInput/ThemedNumberInput.tsx
dhinrichs-scottlogic: if you have aria-disabled you need to add javascript so that you can't use the input. 
readonly already does that and is announced by screen readers. 
the themedTextInput already uses readOnly. We're using aria-disabled on the buttons.

frontend/src/components/DefenceBox/DefenceMechanism.tsx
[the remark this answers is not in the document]
pmarsh-scottlogic: ah, oh well! I didn't think I saw it in the 'before' part of the diff, but I only glanced quickly :)

frontend/src/components/DefenceBox/DefenceMechanism.css
dhinrichs-scottlogic: the article where I found the accessible toggle button says this about ems:
"All sizing is done in ems so these will scale with the surrounding text."

frontend/src/components/DefenceBox/DefenceMechanism.tsx
dhinrichs-scottlogic: feel free to have a play around with it and see if you can find an alternative, but this part was taken from the example code, link is in the css

frontend/src/components/DefenceBox/DefenceMechanism.css
pmarsh-scottlogic: fair enough :)

frontend/src/components/DefenceBox/DefenceMechanism.css
pmarsh-scottlogic: very fair, not strictly relevant to the PR. Happy to leave as is, but would prefer no comment.

frontend/src/components/DefenceBox/DefenceMechanism.tsx
pmarsh-scottlogic: The alternative is to remove the `<form>` element and put the styles onto the `<div>` instead.

image

image

looks and works the same way.

As I understand `<form>`s are for (like their namesake) grouping user input elements together that will all be submitted and processed at once. So I expect the example in [the link](https://adrianroselli.com/2019/03/under-engineered-toggles.html) grouped the toggles in a form because that's how they expected it to be used in real life, in a form setting. Given we're not batch-input-and-processing any information, we may as well remove the form, since it is just wrapping a single element.

On that note, in the [updated intro](https://adrianroselli.com/2019/03/under-engineered-toggles.html) of the webpage it mentions the cases where it's better to use a button for the toggle instead of a checkbox, and it seems as though our app meets all three criteria for using a button, and none (possibly one - I don't know what progressively enhancing the control means) of those for using a checkbox. What do you think?

image

pmarsh-scottlogic:
> > ## Focus problems
> > The focus indicators seem to be getting cut off at the edges for the main config `<summary>` elements.
> > not quite sure which bit of css is causing this.
> 
> @pmarsh-scottlogic Check that you're using the most up-to-date version of this branch. Yesterday I rebased off dev, which had the new focus indication merged in, and I don't have that focus issue.

Ah, I was on the branch `725-defence-toggle-switches-arent-read-out-by-nvda` rather than `try_725_again`. All looks good here :)

frontend/src/components/DefenceBox/DefenceMechanism.css
```diff
@@ -1,106 +1,164 @@
...
 	border-radius: 50%;
+	background: var(--main-text-colour);
+	background-position: center center;
 }
 
-.defence-mechanism .info-box {
-	font-size: 0.875rem;
+.toggles [type='checkbox']:focus + label::before,
+.toggles [type='checkbox']:hover + label::before {
+	box-shadow: 0 0 0.3em var(--main-text-colour);
```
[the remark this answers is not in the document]
pmarsh-scottlogic: Have you noticed that the new focus on the toggles has a fun little animation on it? Pretty neat, did you do that or did it arise out of adapting the shadow style?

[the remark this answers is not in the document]
dhinrichs-scottlogic: I'm not sure what you mean by fun little animation, but I have added the outline to be clearly visible 
image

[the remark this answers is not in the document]
pmarsh-scottlogic: ah, this isn't a problem. The focus looks good! For me it looks like the ring appears and then grows up a little, which I enjoyed so I thought I'd point it out

frontend/src/components/DefenceBox/DefenceMechanism.css
pmarsh-scottlogic: We've had a discussion on Teams for this and agreed it's cool as is

pmarsh-scottlogic (approved):

PR #751, opened 2024-01-09
598 better tooltips

gsproston-scottlogic:
## Description

No longer using `title` on buttons to show tooltips. We were seeing issues where if the button was focused, the tooltip would either be shown nowhere near the button (on Edge) or not shown at all (on Firefox).
Now, we're using a `div`, as recommended [here](https://inclusive-components.design/tooltips-toggletips/).

## Screenshots

Level select button:

image

Defence config reset button:

image

Sending button:

image

## Notes

- Removed the LevelButton component entirely. Instead, we're now using a ThemedButton with a set class name.

## Concerns

- The parent of a ThemedButton with a tooltip must pass an ID to the ThemedButton component which is used to uniquely identify the tooltip. This is so the `button` can use `aria-describedby` and link to the tooltip `div`. I'm not sure how much I like this. We can't guarantee that these IDs will always be unique. We could instead let ThemedButton compute a random unique ID, like a UUID, but I'm not sure of the performance ramifications of that 🤔 
- For the "View Documents" and "send" buttons, we have to specify some `.themed-button` styling outside of ThemedButton.css in order for them to stretch the full width. I'd like to avoid this, but it was the cleanest solution I could think of.

## Checklist

Have you done the following?

- [ ] Added tests


frontend/src/components/ThemedButtons/ThemedButton.tsx
```diff
@@ -28,16 +31,27 @@ function ThemedButton({
 		disabled: ariaDisabled,
 	});
 
+	const tooltipId = `themed-button-desc-${tooltip?.id
```
chriswilty, line 34: Somewhat confused by why the tooltip div needs an id.

gsproston-scottlogic: I think we may have covered this in a daily, but for prosperity's sake, this ID links the `aria-describedby` on the `button` to the tooltip `div`.

pmarsh-scottlogic (approved):
seems pretty good and behaves nicely!

heatherlogan-scottlogic (approved):
Looks/works good to me :)

PR #753, opened 2024-01-10
703 add random sequence enclosure defence back in 

heatherlogan-scottlogic:
## Description

Reword the descriptions on XML tagging and Random sequence enclosure. Also changes prompt config display name to pre-prompt (the ID is still prompt so the config setting can be reused).

Changed the name param on the radio buttons as before, NDVA on firefox was reading out "button checked 1 of 1" instead of "2 of 1". 

Moved the radio buttons into the defence box folder as mentioned previously that ThemedButtons should only be for reused buttons 

## Screenshots
image
image

## Concerns

- Before this change, we could tab through the radio buttons and click space to select. This change means that we can only tab onto the first None button, then navigate by pressing up/down. Thought this was a bug at first but it is the same way as on the Gov website https://design-system.service.gov.uk/components/radios/, so i'm guessing this is standard?...

## Checklist

Have you done the following?

- [ ] Added tests


gsproston-scottlogic (changes requested):
These particular changes all look good to me, but I'm seeing a related bug.

If you activate prompt enclosure and set it to XML tagging, for example. If you then refresh the page, then it will be set back to 'none', but still be active.

frontend/src/components/DefenceBox/PromptEnclosureDefenceMechanism.test.tsx
```diff
@@ -155,12 +147,13 @@ describe('PromptEnclosureDefenceMechanism component tests', () => {
...
 	test('when xml tagging radio button is selected, it is toggled on and others are toggled off', async () => {
 		const newMockDefences = [...mockDefences];
+		newMockDefences[0].isActive = false;
 		newMockDefences[1].isActive = true;
 
 		const { user } = renderComponent(newMockDefences);
 		const radioButton = screen.getByRole('radio', {
 			name: 'xml tagging',
-		});
+		}) as Element;
```
chriswilty, line 156: Why?

frontend/src/components/DefenceBox/PromptEnclosureDefenceMechanism.tsx
```diff
@@ -22,13 +22,20 @@ function PromptEnclosureDefenceMechanism({
...
 	const selectedDefence = getDefence(selectedRadio as DEFENCE_ID);
 
 	function isRadioSelected(radioValue: string) {
 		return radioValue === selectedRadio;
 	}
 
+	function getActiveDefence(): string {
+		const activeDefence = defences.find((defence) => defence.isActive);
+		return activeDefence ? activeDefence.id : 'none';
+	}
```
chriswilty, line 37:
```
return defences.find((defence) => defence.isActive)?.id ?? 'none';
```

chriswilty (commented):
One question, one suggestion, else good to go!

frontend/src/components/DefenceBox/PromptEnclosureDefenceMechanism.test.tsx
heatherlogan-scottlogic: I had used it in the other tests as think i was getting something like a Element vs HTMLElement warning on the user button click

gsproston-scottlogic (approved):
All works for me! Nice one :D

frontend/src/components/DefenceBox/PromptEnclosureDefenceMechanism.test.tsx
chriswilty: I'll remove it, then we're good to go.

chriswilty (approved):
All good now.

Issue #713, opened 2023-12-21
Show error in the handbook if we fail to get system roles


gsproston-scottlogic:
# Bug report

## Description

Show an error on the 'system role' page if we fail to get the system roles from the backend.

## Reproduction steps

Steps to reproduce the behaviour:

1. Don't run the backend
2. Enter story mode (not sandbox)
3. Open the handbook
4. Go to the 'system role' page

## Expected behaviour

Show an error saying we failed to retrieve the system roles.

## Screenshots

Current:

![image](https://github.com/ScottLogic/prompt-injection/assets/103250539/e4965860-91b6-4da7-863f-3e9ce4219505)

Figma design:

![image](https://github.com/ScottLogic/prompt-injection/assets/103250539/9800315a-aca8-4243-ada0-da4ef2847dc6)

## Software (please complete the following information):

- Browser: Edge

## Additional context

[Figma](https://www.figma.com/file/dntI3xrEI40De9KDCMCmMI/AI-prompt-injection?type=design&node-id=765-1945&mode=dev).

Questions:

- We could hard code system roles on the frontend, to avoid having to get them from the backend. The only trouble here would be keeping them in sync if they were ever to change. **- don't bother with this. Too much to worry about for such a niche issue.**

## Acceptance criteria

**GIVEN** the backend isn't running
**WHEN** the handbook is opened to the system role page
**THEN** an error is shown

dcottle-scottlogic:
Hey would I be possible for me to work on this issue?

gsproston-scottlogic:
Hi @dcottle-scottlogic , that'd be great thanks, I'll assign you to this now.

PR #755, opened 2024-01-10
713 system role error message

dcottle-scottlogic:
## Description

Added an error message onto the Systems admin page of the handbook
## Screenshots

**When the backend is not running (error message)**
image

**When the backend is running (no error message)**
image

## Notes

-Updated a couple Theme.css lines to reduced remove a repeated value which I was also going to be working with
-Resolves this issue #713 

## Concerns

- I added this only on the HandbookSystemRole component as I was not sure if it needed to be its own component or not. 

## Checklist

Have you done the following?

- [x] Added tests

### DIFF
--- a/frontend/src/components/HandbookOverlay/HandbookPage.css
+++ b/frontend/src/components/HandbookOverlay/HandbookPage.css
@@ -36,8 +36,8 @@
 }
 
 .system-role-error-message {
-	background-color: var(--error-background-colour);
 	border: 0.125rem dashed var(--error-colour);
-	border-radius: 0.625rem;
+	background-color: var(--error-background-colour);
 	padding: 0.625rem;
+	border-radius: 0.625rem;
 }
--- a/frontend/src/components/HandbookOverlay/HandbookPage.css
+++ b/frontend/src/components/HandbookOverlay/HandbookPage.css
@@ -34,3 +34,10 @@
 	background: var(--handbook-locked-section-colour);
 	text-align: center;
 }
+
+.error-box-system-role {
+	background-color: var(--error-background-colour);
+	border: 0.125rem dashed var(--error-colour);
+	border-radius: 0.625rem;
+	padding: 0.625rem;
+}
--- a/frontend/src/components/HandbookOverlay/HandbookPage.css
+++ b/frontend/src/components/HandbookOverlay/HandbookPage.css
@@ -24,7 +24,7 @@
 	margin: 0;
 }
 
-.role-locked {
+.handbook-terms .role-locked {
 	display: flex;
 	justify-content: center;
 	align-items: center;
@@ -35,7 +35,7 @@
 	text-align: center;
 }
 
-.system-role-error-message {
+.handbook-terms .system-role-error-message {
 	border: 0.125rem dashed var(--error-colour);
 	background-color: var(--error-background-colour);
 	padding: 0.625rem;
--- a/frontend/src/components/HandbookOverlay/HandbookPage.css
+++ b/frontend/src/components/HandbookOverlay/HandbookPage.css
@@ -35,7 +35,7 @@
 	text-align: center;
 }
 
-.error-box-system-role {
+.system-role-error-message {
 	background-color: var(--error-background-colour);
 	border: 0.125rem dashed var(--error-colour);
 	border-radius: 0.625rem;
--- a/frontend/src/components/HandbookOverlay/HandbookSystemRole.test.tsx
+++ b/frontend/src/components/HandbookOverlay/HandbookSystemRole.test.tsx
@@ -32,9 +32,9 @@ describe('HandbookSystemRole component tests', () => {
 		expect(screen.getByText('System Roles')).toBeInTheDocument();
 
 		// make sure no system roles are displayed on the page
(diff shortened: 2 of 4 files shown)


frontend/src/components/HandbookOverlay/HandbookSystemRole.test.tsx
```diff
@@ -121,4 +124,41 @@ describe('HandbookSystemRole component tests', () => {
...
+				systemRoles={emptySystemRoles}
+			/>
+		);
+
+		expect(screen.getByText('System Roles')).toBeInTheDocument();
+
+		// make sure no system roles are displayed on the page
+		expect(screen.queryAllByText(level1SystemRole)).toHaveLength(0);
+		expect(screen.queryAllByText(level2SystemRole)).toHaveLength(0);
+		expect(screen.queryAllByText(level3SystemRole)).toHaveLength(0);
```
chriswilty, line 143: We don't expect these to occur more than once, plus, we can do better than queryByText because these are role="term":

image

So use this instead:

```
expect(screen.queryByRole('term', { name: level1SystemRole })).not.toBeInTheDocument();
```

frontend/src/components/HandbookOverlay/HandbookSystemRole.test.tsx
```diff
@@ -121,4 +124,41 @@ describe('HandbookSystemRole component tests', () => {
...
+
+		// check that the error message is displayed
+		expect(screen.queryAllByText(errorMessage)).toHaveLength(1);
```
chriswilty, line 146: Prefer jest-dom matchers, and prefer single-match calls when you expect none or one match:

```
expect(screen.getByText(errorMessage)).toBeInTheDocument();
```

frontend/src/components/HandbookOverlay/HandbookSystemRole.test.tsx
```diff
@@ -121,4 +124,41 @@ describe('HandbookSystemRole component tests', () => {
...
+		const numLevelsCompleted = 0;
+
+		render(
+			<HandbookSystemRole
+				numCompletedLevels={numLevelsCompleted}
+				systemRoles={systemRoles}
+			/>
+		);
+
+		expect(screen.getByText('System Roles')).toBeInTheDocument();
```
chriswilty, line 159: Strictly speaking you don't need this expect, as it is tested elsewhere.

frontend/src/components/HandbookOverlay/HandbookSystemRole.test.tsx
```diff
@@ -121,4 +124,41 @@ describe('HandbookSystemRole component tests', () => {
...
+
+		// check that the error message is not being displayed
+		expect(screen.queryAllByText(errorMessage)).toHaveLength(0);
```
chriswilty, line 162:
```
expect(screen.queryByText(errorMessage)).not.toBeInTheDocument();
```

frontend/src/components/HandbookOverlay/HandbookPage.css
```diff
@@ -34,3 +34,10 @@
 	background: var(--handbook-locked-section-colour);
 	text-align: center;
 }
+
+.system-role-error-message {
```
chriswilty, line 38: To be safe, I'd prefer this selector to include the handbook-terms class:

```
.handbook-terms .system-role-error-message {
```

Mind you, we already have inconsistency here because we should be doing that for role-locked as well. That's an important one because it's generic enough that it could leak style rules... So could you fix that one here as well please? Thanks!

```
.handbook-terms .term .role-locked {
```

gsproston-scottlogic (approved):
Looks great! Thanks for your work on this 😄

chriswilty (changes requested):
Good work, just a handful of suggestions.

frontend/src/components/HandbookOverlay/HandbookSystemRole.test.tsx
chriswilty: I now see that you copied the existing pattern... so I understand why you did it! But we can improve them all, throughout this file, as suggested.

frontend/src/components/HandbookOverlay/HandbookSystemRole.test.tsx
```diff
@@ -121,4 +124,41 @@ describe('HandbookSystemRole component tests', () => {
 		const lockedBox = container.getElementsByClassName('role-locked');
 		expect(lockedBox).toHaveLength(0);
```
chriswilty, line 125: @dcottle-scottlogic This is also a no-no, testing to see if classnames have been used. Again, not something you added! I wonder if we might pair on this to correct things in this test file. I can explain as we go.

dcottle-scottlogic: @chriswilty Yeah sure can do 👍

chriswilty (commented):
Haha, well looks like you fixed most of them already! I'm a bit snowed under today, but maybe I can ping you tomorrow to chat about the other issues in this test file? I think it's useful info to share with you, even though you didn't author the original file.

chriswilty (approved):
Looks good to me.

Was interesting to learn about the term and definition roles, and how their accessible names are (and are not) acquired 😕

PR #740, opened 2024-01-04
708 move logic for detecting output defence bot filtering

gsproston-scottlogic:
## Description

Describe what your code changes do.

## Screenshots

Where applicable, show screenshots of the visual changes.

## Notes

- Bullet point list with
- any notable code changes
- or explanations regarding this PR

## Concerns

- Bullet point list
- with any questions
- or concerns about this PR

## Checklist

Have you done the following?

- [ ] Linked the relevant Issue
- [ ] Added tests
- [ ] Ensured the workflow steps are passing

PR #758, opened 2024-01-11
707 refactor of chatgptchatcompletion

pmarsh-scottlogic:
## Description

REFACTOR: moves the logic for _setting the system role in chat history_ to its own method (`setSystemRoleInChatHistory`), and then instead of calling that method from `chatGptChatCompletion` (`openAI` module), we call it from `handleChatToGPT` (`chatController` module). The idea of moving where it's invoked is to make methods responsible for fewer things.

Rewrites tests accordingly.

## Notes

- we also alter the logic to reduce mutation. Except we change the value of the chat history in the session once we've made the updated value (with the system role set)
- unit tests the new method `setSystemRoleInChatHistory` on its own, then just checks that it gets called in `handleChatToGPT`
- removes the old integration test that checked setting system roles in the chat history

## Concerns

- Wasn't sure where to put the definition of `setSystemRoleInChatHistory`, so I left it in `openai`. In my ideal word, we'd have some module "chatService" which sits in between the `chatController` and the lower level modules (`openai`, `langchain`) which would deal with this sort of thing.

## Checklist

Have you done the following?

- [x] Added tests


backend/src/openai.ts
```diff
@@ -218,50 +218,55 @@ async function chatGptCallFunction(
...
-		// remove the system role from the chat history
-		while (
-			chatHistory.length > 0 &&
-			chatHistory[0].completion?.role === 'system'
-		) {
-			chatHistory.shift();
-		}
+		return chatHistory.filter(
+			(message) => message.completion?.role !== 'system'
+		);
```
gsproston-scottlogic, line 260: I was going to say this would be less efficient, but a quick google suggests `shift` to be O(N) 🤷 
So overall this reads a lot nicer, AND is likely to be just as efficient, if not more so. Nice one!

gsproston-scottlogic (changes requested):
Very nice! The only thing I'd prefer would be to move `setSystemRoleInChatHistory` out of openai.ts. 
I'd argue it can go in the chat controller, as that's the only place that uses it, although I appreciate that file is getting a little large.
Adding services for each controller would be nice, but a little overkill for this ticket. Could be worth making a new issue for it though? 👀 
Alternatively, in #740 I added a chat utils file, could go there?

backend/src/openai.ts
pmarsh-scottlogic: I'm surprised shift() is O(N)! I guess it moves all the values about in memory so there isn't a dead bit of memory where the first element used to be

backend/test/unit/openai.test.ts
```diff
@@ -69,6 +77,123 @@ describe('openAI unit tests', () => {
...
+	...mockChatHistoryWithoutSystemRole,
+];
+
+const mockIsDefenceActive = jest.fn();
+
+jest.mock('@src/defence', () => ({
+	getSystemRole: jest.fn().mockImplementation(() => mockSystemRolePrompt),
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-return
+	isDefenceActive: jest.fn().mockImplementation(() => mockIsDefenceActive()),
+}));
```
chriswilty, line 124: I find this test suite a bit confusing. We have two top-level describe blocks, sharing some setup/teardown, but then we have a bunch of consts and mocks which are also at the top-level but only used in one or other of the describes.

This might be a symptom of not knowing where some of this logic should sit. If there are functions in the module under test that only use a subset of the imports, then it's ok to stub them at the top level, but then you should mock the behaviour only where it is needed. So here, for example, we wouldn't necessarily want to `mockImplementation` for getSystemRole for tests within a describe that don't actually call getSystemRole.

And finally, if you are mocking the `@src/defence` module, then I believe getSystemRole and isDefenceActive are already empty mocks:

```
import { isDefenceActive } from '@src/defence';

jest.mock('@src/defence');

describe('some function', () => {
  const isDefenceActiveMock = isDefenceActive as jest.Mocked<typeof isDefenceActive>;

  afterEach(() => {
    isDefenceActiveMock.mockReset();
  });

  test('something', () => {
    isDefenceActiveMock.mockImplementationOnce(() => true);
    ...
  });
});
```

More info at https://jestjs.io/docs/mock-function-api#jestmockedsource

chriswilty (changes requested):
Just a few thoughts.

backend/test/integration/chatController.test.ts
```diff
@@ -152,12 +153,17 @@ describe('handleChatToGPT integration tests', () => {
...
+			openaiModule,
+			'setSystemRoleInChatHistory'
+		);
 		mockCreateChatCompletion.mockResolvedValueOnce(
 			chatResponseAssistant('Howdy human!')
 		);
 
 		await handleChatToGPT(req, res);
 
+		expect(mockedSetSystemRoleInChatHistory).toHaveBeenCalled();
```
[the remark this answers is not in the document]
pmarsh-scottlogic: You're right. Actually, we're already checking that the system role is added to the history. Req begins with empty history for level 1, then we check that the system role is the first element later

backend/test/unit/utils/chat.test.ts
pmarsh-scottlogic, line 1: I added the tests for `setSystemRoleInChatHistory` then decided I could split the file apart into `pushMessageToHistory.test.ts` and `setSystemRoleInChatHistory`

backend/test/unit/controller/chatController.test.ts
```diff
@@ -297,6 +324,41 @@ describe('handleChatToGPT unit tests', () => {
 				})
 			);
 		});
+
+		test('GIVEN output filtering defence enabled WHEN handleChatToGPT called THEN it should return 200 and blocked reason', async () => {
```
pmarsh-scottlogic, line 328: this test ain't got anything to do with this ticket, but I noticed it was missing and it's pretty straightforward so I added it in.

backend/test/unit/openai.test.ts
```diff
@@ -1,63 +1,26 @@
...
-		initDocumentVectors: jest.fn(),
-	};
-});
+describe('getValidModelsFromOpenAI', () => {
+	const mockOpenAI = OpenAI as jest.MockedClass<typeof OpenAI>;
+	mockOpenAI.prototype = {
+		models: {
+			list: jest.fn(),
+		},
+	} as unknown as typeof mockOpenAI.prototype;
```
chriswilty, line 15: Don't think this is the correct way to mock a class, you should generally leave prototypes alone. This seems to work:

```
const mockListFn = jest.fn<OpenAI.Models['list']>();
jest.mocked(OpenAI).mockImplementation(() => ({
	models: {
		list: mockListFn,
	},
} as unknown as jest.MockedObject<OpenAI>));
```

Then in your test:

```
mockListFn.mockResolvedValueOnce({
	data: mockModelList,
} as OpenAI.ModelsPage);
```

For the sake of completion, you might want to include an afterEach() to reset mockListFn.

backend/test/unit/utils/chat.ts/pushMessageToHistory.test.ts
```diff
@@ -0,0 +1,132 @@
...
+			chatHistory,
+			generalChatMessage
+		);
+		expect(updatedChatHistory.length).toBe(maxChatHistoryLength);
+		expect(updatedChatHistory[0]).toEqual(systemRoleMessage);
+		expect(updatedChatHistory[updatedChatHistory.length - 1]).toEqual(
+			generalChatMessage
+		);
+	}
+);
```
chriswilty, line 132: Much cleaner ❤️

chriswilty (changes requested):
This is great 👌 

One suggestion for test mocking. Mocking classes is pretty complex.... One very good reason to avoid classes in our own code!

backend/test/unit/openai.test.ts
pmarsh-scottlogic: yup, seems nicer

backend/test/unit/openai.test.ts
```diff
@@ -1,4 +1,5 @@
 import { beforeEach, describe, expect, jest, test } from '@jest/globals';
+import { afterEach } from 'node:test';
```
chriswilty, line 2: Whoops, wrong import...

chriswilty (changes requested):
Just an import to correct and we're all good 👍

Issue #776, opened 2024-01-16
Rename "Instruction Defence" and "Prompt Enclosure"


dhinrichs-scottlogic:
# Bug report

## Description
The "Instruction Defence" name is causing issues, because when we toggle a defence on we use the function addInfoMessage (MainComponent.tsx) which already contains the word "defence" in the string, hence why we see "instruction defence defence activated" in the chatbox.  
Only using "Instruction" might not be very clear. Perhaps something like "Pre-Prompt" would work. Discuss with the team which wording to use before implementing this. 

Further more, the "Prompt Enclosure" heading under which this defence can be found is slightly confusing since the Instruction Defence is not enclosing the prompt, but prepending the prompt, so a different name (e.g "Prompt Transformation") would be better. 
Discuss with the team which wording to use before implementing this. 

## Reproduction steps

Steps to reproduce the behaviour:

1. Go to the "Prompt Enclosure" defence configuration
2. Toggle the "Instruction Defence" on.
3. See error in chat box "instruction defence defence activated"

## Expected behaviour
Clear names, no duplicating words

## Screenshots
![Image](https://github.com/ScottLogic/prompt-injection/assets/125262707/217b9456-5155-448f-9053-ff0ad41854b0)

## Acceptance criteria

**GIVEN** some setup
**WHEN** an action happens
**THEN** the expected outcome

Issue #774, opened 2024-01-15
Give the configurations response error messages


dhinrichs-scottlogic:
# Bug report

## Description
We currently don't have a mechanism to display to the user that the backend call for the configurations failed. 
When the backend isn't responding, the configuration buttons can't be toggled, but there is no indication as to why. 

## Reproduction steps

Steps to reproduce the behaviour:

1. Start the backend and frontend
2. Go to the sandbox and toggle one of the defences (e.g. Character Limit) on. You will see a message in the chatbox saying "character limit defence activated" and in the networks tab of the dev tools you will see the 200 code that your request was successful. 
3. Turn your backend off. Try to turn any of the defences on or off. 
4. Observe that the buttons cannot be toggled, that there is the correct error response in the network, but there is no message in the UI indicating why the buttons can't be toggled, that the backend request wasn't successful. 

## Expected behaviour
Given the user tries to adjust the defences (on/off, changing the input, adjusting the sliders, changing the model)
When the backend request for the configuration update is failing
Then the user receives an error message in the chatbox. 

## Screenshots
![Image](https://github.com/ScottLogic/prompt-injection/assets/125262707/171b7ccc-94dc-47c0-a547-8d9a9968d287)

## Acceptance criteria

**GIVEN** some setup
**WHEN** an action happens
**THEN** the expected outcome

Issue #742, opened 2024-01-05
Model configuration sliders update on every step


gsproston-scottlogic:
# Bug report

## Description

When the user drags the model configuration slider, the frontend sends a request to the backend for every step, not just one for when the user stops dragging.

## Reproduction steps

Steps to reproduce the behaviour:

1. Go to sandbox mode
2. Open browser dev tools to the network tab
3. Drag around a slider in model configuration 
4. See that a request is sent for each step of the slider

## Expected behaviour

Just one request is sent once the user stops dragging.

## Screenshots

Bunch of requests just from dragging a slider once:

![image](https://github.com/ScottLogic/prompt-injection/assets/103250539/3b8f1fbe-5116-4815-b437-e48c5a16b22c)

## Software (please complete the following information):

- Browser: Edge

## Acceptance criteria

**WHEN** the user is dragging a model configuration slider
**THEN** requests are NOT sent until the user stops dragging

**GIVEN** the user is dragging a model configuration slider
**WHEN** the user stops dragging
**AND** the value is different from the original one
**THEN** a single request is sent 

**GIVEN** the user is dragging a model configuration slider
**WHEN** the user stops dragging
**AND** the value is the same as the original one
**THEN** no requests are sent

Issue #741, opened 2024-01-05
Transformed messages not showing correctly 


gsproston-scottlogic:
# Bug report

## Description

After refreshing the page or switching levels, the user's transformed messages don't show correctly. The original message is no longer bold and the info message isn't shown. 

## Reproduction steps

Steps to reproduce the behaviour:

1. Go to sandbox mode
2. Activate XML tagging
3. Send a chat message
4. Wait for the reply
5. See that there's an info message, and the original message is bold within the transformed message
6. Refresh the page OR switch levels
7. No info message and the original message is not bold.

## Expected behaviour

The transformed message should contain the original message in bold, and should be below an info message saying the original message has been transformed.

## Screenshots

Before refreshing or switching level:

![image](https://github.com/ScottLogic/prompt-injection/assets/103250539/8f4d95fa-dadb-476c-8c56-16dae529bfb7)

After:

![image](https://github.com/ScottLogic/prompt-injection/assets/103250539/081ce6ce-f5a7-47a0-a14d-d516683bd8cd)

## Software (please complete the following information):

- Browser: Edge

## Acceptance criteria

**GIVEN** a message has been transformed
**WHEN** the page is refreshed
**THEN** the transformed message contains the original message **in bold**
**AND** there is still an info message above the transformed message explaining that the original message has been transformed

**GIVEN** a message has been transformed
**WHEN** the user switches level and then back to the original level
**THEN** the transformed message contains the original message **in bold**
**AND** there is still an info message above the transformed message explaining that the original message has been transformed

pmarsh-scottlogic:
# getChatHistory

```typescript
function getChatHistory {
  // get chatHistory from backend
  // for each message in retrieved chat history (type `ChatHistoryMessage`), convert it to type `ChatMessage` and add it to `chatMessages` array
  return chatMessages
}
```
this could be refactored to use a .filter (to filter out the messages of type `SYSTEM` and `FUNCTION_CALL`) and a .map to convert each message from type `ChatHistoryMessage` to `ChatMessage`. Also these could have more distinct names.

`getChatHistory` is called by `setNewLevel`, which gets the chat history and then updates the frontend `messages` state variable.

`setNewLevel` is only called in place: in a useEffect.
```typescript
	useEffect(() => {
		void setNewLevel(currentLevel);
	}, [currentLevel]);
```

pmarsh-scottlogic:
when we send a chat message with xml tagging we want
[...]

PR #778, opened 2024-01-16
715 defence successfully configured message should not disappear after a delay

dhinrichs-scottlogic:
## Description
Previously, when a user changed or reset the text or number inputs of the defence and model configurations, a message would pop up for 3 seconds to indicate to the user that the change has been made. This can easily be missed, which is why I moved this message to the chat, where the user already gets updates when defences get toggled on/off. 
Additionally, I added an info message for the GPT model selection and the model sliders 

### Before:
image

### After:
image

## Notes
- changed two key/value pairs of the enum DEFENCE_ID: FILTER_USER_INPUT to INPUT_FILTERING and FILTER_BOT_OUTPUT to OUTPUT_FILTERING
- added the info messages to the chatHistory
- removed the old pop up info messages

## Concerns
- Due to the naming of the Information Defence, "defence" is repeated in the chat, there is already [a ticket](https://github.com/ScottLogic/prompt-injection/issues/776) to address this.
- There is currently no error messages for the user if the backend fails while we are changing a configuration. There's a [separate ticket ](https://github.com/ScottLogic/prompt-injection/issues/774)to address this. 
- The defence.name and the DEFENCE_ID are slightly different q/a LLM versus qa LLM. I couldn't change the enum to include the "/" in the enum key (which is what's being used in the function to update the config) and I didn't want to create a special conditional to change the output to the UI. Personally I think the difference between q/a and qa is acceptable for now. There is a general question about how useful the DEFENCE_ID enum is given that the defence names and the enum is almost identical, so this could become a bigger refactor. 

## Checklist

Have you done the following?

- [ ] Added tests


pmarsh-scottlogic:
this change interacts amusingly with this bug #742 
Not something to block this PR, but we should prioritise the bug once this is in

pmarsh-scottlogic (changes requested):
I've not looked at the code yet, but this is a great change!

First thing I notice is that on refresh, or changing level then back again, these lovely new messages vanish! That's because when we add an info message to the messages state object, we must also update the chatHistory in the backend. See `addInfoMessage` defined in _MainComponent.tsx_.
Almost the same problem as #741 

I'll do a proper review next week

Edit: Turns out `addInfoMessage` in MainComponenet.tsx will add the message to the front and as well as the backend history. So the fix is to use `addInfoMessage` rather than `addChatMessage`.

frontend/src/components/MainComponent/MainComponent.tsx
```diff
@@ -184,6 +184,9 @@ function MainComponent({
 			return defence;
 		});
 		setDefencesToShow(newDefences);
+		// add info message to chat
```
pmarsh-scottlogic, line 187: I'd prefer no comment here. I think the code reads well already without it!

frontend/src/components/MainComponent/MainComponent.tsx
```diff
@@ -216,6 +219,9 @@ function MainComponent({
 				return defence;
 			});
 			setDefencesToShow(newDefences);
+			// add info message to chat
```
pmarsh-scottlogic, line 222: same here, don't think the comment is needed

pmarsh-scottlogic (changes requested):
Super sleek!

Only some minor things from me

frontend/src/components/ModelBox/ModelConfigurationSlider.tsx
```diff
@@ -2,22 +2,32 @@ import { Slider } from '@mui/material';
...
 	config: CustomChatModelConfiguration;
+	addChatMessage: (message: ChatMessage) => void;
 }) {
 	const [value, setValue] = useState<number>(config.value);
 	const [showInfo, setShowInfo] = useState<boolean>(false);
 
 	async function handleValueChange(_: Event, value: number | number[]) {
 		const val = Array.isArray(value) ? value[0] : value;
 		setValue(val);
+		addChatMessage({
```
pmarsh-scottlogic, line 27: You should import and use `addInfoMessage` instead. That'll sort out the refresh problem

frontend/src/components/ModelBox/ModelSelection.tsx
```diff
@@ -22,6 +29,10 @@ function ModelSelection({ chatModelOptions }: { chatModelOptions: string[] }) {
 		if (!isSettingModel && selectedModel) {
 			const currentSelectedModel = selectedModel;
 			console.log(`selected model: ${currentSelectedModel}`);
+			addChatMessage({
+				type: CHAT_MESSAGE_TYPE.INFO,
+				message: `GPT model changed to: ${currentSelectedModel}`,
+			});
```
pmarsh-scottlogic, line 35: You should import and use `addInfoMessage` instead. That'll sort out the refresh problem

frontend/src/components/MainComponent/MainComponent.tsx
```diff
@@ -184,6 +184,9 @@ function MainComponent({
 			return defence;
 		});
 		setDefencesToShow(newDefences);
+		// add info message to chat
+		const displayedDefenceId = defenceId.replace(/_/g, ' ').toLowerCase();
+		addInfoMessage(`${displayedDefenceId} defence reset`);
```
chriswilty, line 189: You have these two lines duplicated now; may as well pull out into a separate function, which takes param 'reset' or 'configured' for the message to add.

chriswilty (changes requested):
One primary concern, but all else good.

I assume there are no tests covering these components? If you want to learn some Testing Library style component tests, you might consider adding some for one of the simpler components, or some around the proposed validation changes. Doesn't need to be in this PR though.

dhinrichs-scottlogic:
This all turned out much more complicated than anticipated, so after some discussion decided to approach it slightly differently and started again from scratch. 
New PR: https://github.com/ScottLogic/prompt-injection/pull/802

PR #788, opened 2024-01-19
Fix reset progress bug in remote

### LINKED ISSUE #787: Reset button not working in remote deployment
# Bug report

## Description

It was noted in demo that the Reset Progress button is not working when deployed to remote. Seems this is because of a leading slash on the path the frontend uses to send the request - see frontend/src/service/levelService and compare to e.g. 

## Reproduction steps

Steps to reproduce the behaviour:

0. Deploy live env
1. Open the Network tab in browser devtools
2. Click on Reset All Progress button in UI
3. Click Reset in the confirmation dialog. Note that the dialog does not close.
4. Note error response in Network tab.

You will see that the path is different for this request, missing the "/dev" prefix compared with other successful requests. It manifests as a CORS issue but that's simply because the path was not recognised, so the request didn't even reach the backend service.

## Expected behaviour

All progress is reset, dialog closes, and no error seen in Network tab.

## Screenshots

Image

## Software (please complete the following information):

- all browsers

## Acceptance criteria

**GIVEN** the application is deployed in AWS
**AND** I have begun a chat with the bot
**WHEN** I click on Reset Progress button in header
**AND** a confirmation dialog opens
**AND** I click Reset button
**THEN** the dialog closes
**AND** all chats and level progress are reset to the start of the game

chriswilty:
## Description

The relevant change is removing the leading slash from the request path in levelService. The other change is just for consistency, to match the other services.

Resolves #787 

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/frontend/src/service/documentService.ts
+++ b/frontend/src/service/documentService.ts
@@ -2,14 +2,15 @@ import { DocumentMeta } from '@src/models/document';
 
 import { getBackendUrl, sendRequest } from './backendService';
 
+const PATH = 'documents';
+
 async function getDocumentMetas(signal?: AbortSignal): Promise<DocumentMeta[]> {
-	const path = 'documents';
-	const response = await sendRequest(path, { method: 'GET', signal });
+	const response = await sendRequest(PATH, { method: 'GET', signal });
 	let documentMetas = (await response.json()) as DocumentMeta[];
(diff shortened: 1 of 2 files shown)


pmarsh-scottlogic (approved):
sweet

PR #790, opened 2024-01-19
789 Reduce UI bundle size

### LINKED ISSUE #789: Reduce UI bundle size, by reducing logo sizes and removing unused fonts
Two things can dramatically reduce our UI bundle size, which is currently above 20MB:

Image

- Do not register japanese and chinese fonts in ExportContent.tsx, as they aren't even being used; if we do that then they won't be included in the bundle although for now they can stay in our repo
- Reduce the size of the spylogic logo, by using smaller-sized PDFs - Mark has put these in our project google drive.

chriswilty:
## Description

Removes unused font files and dramatically reduces size of spylogic logo files.

## Screenshots

Before:
298111561-7e814375-bfdf-4d83-9a1b-0b9b92e6efd2

After:
image

Approx 25MB removed 🎉 

Resolves #789 

## Checklist

Have you done the following?

- [ ] Added tests


frontend/src/assets/images/SpyLogicLogo_Negative.png
pmarsh-scottlogic, line 1: I don't think this is used anywhere, nor is it planned to be. Remove?

frontend/src/assets/fonts/README.md
```diff
@@ -1,19 +1,25 @@
...
+- NotoSansHebrew-Regular.ttf
 - NotoSansThai-Regular.ttf
 - NotoSerifBengali_Condensed-Regular.ttf
+- NotoSerifDevanagari-Regular.ttf
 - NotoSerifGurmukhi-Regular.ttf
-- NotoSansHebrew-Regular.ttf
+
+Japanese and Chinese font packs are not included in the CombinedFonts family as they are prohibitively large, so if you are
+running the application locally and wish to export chats in either of those fonts, you will need to modify the code in
+[ExportContent.tsx](../../components/ExportChat/ExportContent.tsx) accordingly.
```
pmarsh-scottlogic, line 22: worth mentioning explicitly that these will need to be downloaded? A user might assume they're in the files somewhere given that there's commented out code that mentions them.

frontend/src/components/ExportChat/ExportContent.tsx
```diff
@@ -8,21 +8,22 @@ import {
...
 import ExportChatBox from './ExportChatBox';
 import ExportEmailBox from './ExportEmailBox';
 
+/*
+Chinese and Japanese font support: currently we're using a Combined font pack,
+but these two font packs are huge so they're kept separate. Only register one
+font pack at a time, then set fontFamily to match in the stylesheet beneath.
+*/
+//Font.register({ family: 'NotoSerifJP', src: NotoSerifJP });
+//Font.register({ family: 'NotoSansSC', src: NotoSansSC });
```
pmarsh-scottlogic, line 24: Actually, if we're going to leave this example code in, then maybe worth doubling down and mentioning explicitly in the comment above that the necessary files must be downloaded.

frontend/src/assets/fonts/README.md
```diff
@@ -1,19 +1,25 @@
...
-- NotoSerifDevanagari-Regular.ttf
 - NotoKufiArabic-Regular.ttf
+- NotoSansHebrew-Regular.ttf
 - NotoSansThai-Regular.ttf
 - NotoSerifBengali_Condensed-Regular.ttf
+- NotoSerifDevanagari-Regular.ttf
 - NotoSerifGurmukhi-Regular.ttf
-- NotoSansHebrew-Regular.ttf
+
+Japanese and Chinese font packs are not included in the CombinedFonts family as they are prohibitively large, so if you are
```
pmarsh-scottlogic, line 20: Seems weird to mention Japanese and Chinese specifically. We're missing korean too, non? I suppose Japanese and Chinese might be more likely to come up...

pmarsh-scottlogic (changes requested):
Why is 6 scared of 7?
Because `#789` 

Anyway, I've got some confusions which you might be able to clear up.
As I understand, we've packed together a bunch of fonts for different scripts into CombinedFonts.ttf. We've listed what fonts are in there, all good. Now, we've deleted two large packs that aren't included within CombinedFonts.ttf. Fair enough.

My question is, when are we ever going to use any of those fonts in the font file that aren't CombinedFonts.ttf? It seems weird that we're treating these languages and files inconsistently. It would seem easier to go like: "We have one font file CombinedFonts.ttf which is is a bundle of the following font files ... If you require font support beyond this, then download it from the website ...". Then we can delete all the fonts apart from CombinedFonts.ttf. It seems weird to keep the ones that are already merged if we're not going to use them anywhere else.

If we go with the above approach then you can probably dismiss my individual suggestions about code comments. My overall concern is if I was a user that wanted my exports in Japanese, I don't think it's immediately clear what steps I should take to make that happen.

you get me?

frontend/src/assets/fonts/README.md
chriswilty: Only if we decide to remove those two fonts from the repo 😉

frontend/src/components/ExportChat/ExportContent.tsx
chriswilty: This is for running locally, so unless we remove the fonts from the repo, they don't need to be downloaded.

frontend/src/assets/fonts/README.md
chriswilty: I know, it's all a bit random to me, but I think it was a work-in-progress that just got kicked into the long grass.

I would be tempted to do as you suggest, and remove the chinese / japanese font files from the repo, plus the commented out lines, and augment the README to say how to use different export fonts.

frontend/src/assets/fonts/README.md
chriswilty: Yep, done.

frontend/src/components/ExportChat/ExportContent.tsx
chriswilty: Fonts now removed, so README and comment updated.

frontend/src/assets/fonts/README.md
chriswilty: Done.

chriswilty:
@pmarsh-scottlogic Comments addressed, files removed. Another pass please?

pmarsh-scottlogic (approved):
yup, lovely

Issue #658, opened 2023-12-13
Sporadically failing UI tests


chriswilty:
# Bug report

## Description

Sporadic test failures, likely due to requests not being cancelled on unmount.

## Reproduction steps

Steps to reproduce the behaviour:

1. Run the DocumentViewBox.test.tsx suite a few times ...
2. Keep running it until you see the "next button is clicked" test fail.

## Expected behaviour

Tests do not fail.

## Screenshots

Below you see the expected change from document 1 to 2 after clicking the Next button, but then it reverts to document 1 again:

![Image](https://github.com/ScottLogic/prompt-injection/assets/15246391/81cf2689-2c73-42af-bbb2-86e3460eb530)

However, it is a bit odd that the log statement occurs more than once, so some additional investigation is needed.

## Software (please complete the following information):

- Browser: None, but uses HappyDom virtual DOM, via vitest tests.

## Additional context

React in _strict mode_ performs a double-render when running in dev mode, which I think might be causing this. If we are not cancelling the original request to fetch the document metas, then they will be set twice, and the second time the document viewer will likely be reset to display the first document. If the Next button click occurs between those two events, and the test runs fast enough, then we have a race condition in which the `expect` might be invoked before or after that second reset event.

## Acceptance criteria

Tests pass consistently.... Not easy to verify!

But no testing needed. tests don't need testing

pmarsh-scottlogic:
difficult to test because the whole suite is skipped! Which, to be fair, solves the sporadic failure problem....

chriswilty:
Hmm, the whole suite is skipped? That sounds wrong :(

@dhinrichs-scottlogic to take a quick look... Thanks!

dhinrichs-scottlogic:
> Hmm, the whole suite is skipped? That sounds wrong :(
> 
> @dhinrichs-scottlogic to take a quick look... Thanks!

was a quick fix we agreed on before Xmas because the DocViewer caused the test to fail. Need to find a workaround.

pmarsh-scottlogic:
## My understanding of the problem

One of the flaky tests is this:
```typescript
test('WHEN the next button is clicked THEN the next document is shown', async () => {
			const { user } = renderDocumentViewBox();
			// wait for header to load
			await screen.findByText(documents[0].filename);

			await user.click(getNextButton());

			expect(
				await screen.findByText(documents[1].filename)
			).toBeInTheDocument();
			expect(
				screen.getByText(`2 out of ${documents.length}`)
[...]

PR #795, opened 2024-01-22
658 Mock the documentViewer in unit tests

pmarsh-scottlogic:
## Description

fixes DocumentViewBox.test.tsx

## Notes

- unskips the test file
- mocks away the library component `<DocViewer>`
- tests that we're rendering the right document by checking that the `<DocViewer>` is called with the correct args, rather than checking that the `<DocViewer>` is correctly rendering stuff on the page.

## Concerns

- We are neglecting to test that `<DocViewer>` is called with the correct `config` and `renderer`. When we tested with both of those I was getting a [weird issue](https://github.com/ScottLogic/prompt-injection/issues/658#issuecomment-1886814008) with the `expect` statement. I think they are not crucial to test, however.

## Checklist

Have you done the following?

- [x] Added tests (well, unskipped them!)

### DIFF
--- a/frontend/src/components/DocumentViewer/DocumentViewBox.test.tsx
+++ b/frontend/src/components/DocumentViewer/DocumentViewBox.test.tsx
@@ -142,16 +142,12 @@ describe('DocumentViewBox component tests', () => {
 			expect(
 				screen.getByText(`1 out of ${documents.length}`)
 			).toBeInTheDocument();
-			expect(mockDocumentViewer).toHaveBeenLastCalledWith({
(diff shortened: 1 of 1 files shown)


chriswilty (approved):
Tests all pass for me locally 👍 

Mocking components can be a bit ugly, as you've found out, but I think the spanner in the works here was that DocViewer is both a memoised component and a ref-forwarding component. My usual mocking tricks didn't seem to work - I tried!

Nice job anyway, and I'm not fussed about the props we don't really care about.

PR #796, opened 2024-01-22
742 stops model config slider updating on every step

pmarsh-scottlogic:
## Notes

- instead of submitting the new value on every `onChange` event, we now update the component's `value` state variable on every `onChange` event, and only submit it on the `onChangeCommitted` event.

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/frontend/src/components/ModelBox/ModelConfigurationSlider.tsx
+++ b/frontend/src/components/ModelBox/ModelConfigurationSlider.tsx
@@ -1,5 +1,5 @@
 import { Slider } from '@mui/material';
-import { useEffect, useState } from 'react';
+import { SyntheticEvent, useEffect, useState } from 'react';
 import { AiOutlineInfoCircle } from 'react-icons/ai';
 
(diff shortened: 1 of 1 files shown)


chriswilty (approved):
I note that onChangeCommitted only fires on mouseup, not on (arrow) keyup, so if a keyboard user holds down an arrow key, we still see multiple events fired.

That's just a MUI Slider implementation choice though - the HTML spec does not state when the change event should fire for a native `<input type="range">`, neither for mouse nor keyboard events, so MUI don't have any native spec to copy.

pmarsh-scottlogic:
> I note that onChangeCommitted only fires on mouseup, not on (arrow) keyup, so if a keyboard user holds down an arrow key, we still see multiple events fired.
> 
> That's just a MUI Slider implementation choice though - the HTML spec does not state when the change event should fire for a native `<input type="range">`, neither for mouse nor keyboard events, so MUI don't have any native spec to copy.

the only way I can see to fix that is to have a button to press when we're done configuring...

PR #800, opened 2024-01-25
741 transformed messages not showing correctly

pmarsh-scottlogic:
## Description

Fixes a bug where transformed messages would appear differently after page refresh. A user's transformed message stays shown with the original message in bold,  and the info message "xml tagging enabled, your message has been transformed" remains.

## Screenshots

image

### main ticket changes

- We have the type `transformedMessage` which keeps a pre-message, message and post-message string, allowing us to render those parts of the string differently in the chat feed. The backend chat history now stores an optional `transformedMessage` of this type, which is returned to the frontend for rendering. This fixes the transformed message appearing differently after refresh.
- To fix the info message disappearing, we now add the info message to the backend's chat history as part of the logic which does the transformation to the user's message in the backend. To get the info message to the frontend for rendering in the chat box feed, we return it as a new property in `ChatHttpResponse`

### testing
- New unit test added to chatController.test which checks that the history is updated properly with the transformed message and info message.
- Refactors the unit tests for transforming messages and moves them into their own file.

### refactoring
- combines two if statemets in `processChatResponse` which were both checking if we had met the win condition.
- renames `incrementNumCompletedLevels` to `updateNumCompletedLevels` (this is called when we win a level. Renamed it because we don't increment if the beaten level has been beaten before.)
- renames `ChatHistoryMessage` to `ChatMessageDTO` (this is the object that is returned when the frontend requests the entire chat history)
- refactors `getChatHistory` method in frontend's chatService.ts just to make it read a bit more clearly.
- `transformedMessage`, `transformedMessageInfo` and `transformedMessageCombined` are now packaged under one object: `messageTransformation`. It is an object of this type, which is now returned by `transformMessage`
- We remove the `isTriggered` property from `Defence` because it wasn't being used anymore (replaced by defenceReport)

## Concerns

- Happy to move some refactorings to a separate PR if requested.

## Checklist

Have you done the following?

- [ ] Added tests


backend/src/controller/chatController.ts
```diff
@@ -43,28 +43,30 @@ function combineChatDefenceReports(
...
+				infoMessage: messageTransformation.transformedMessageInfo,
+			},
 			{
 				completion: {
 					role: 'user',
-					content: transformedMessage,
+					content: messageTransformation.transformedMessageCombined,
 				},
 				chatMessageType: CHAT_MESSAGE_TYPE.USER_TRANSFORMED,
+				transformedMessage: messageTransformation.transformedMessage,
```
chriswilty, line 66: I'm a little unsure about the types now, they feel a little loose. Here we only expect transformedMessage if the type is USER_TRANSFORMED, right? Additionally, completion.role will always be 'user' in that case.

Seems maybe we could split our types for a touch more type-safety, and make ChatHistoryMessage a union type of these.

```
type ChatUserMessageTransformed = {
	completion: ChatCompletionUserMessageParam;
	chatMessageType: CHAT_MESSAGE_TYPE.USER_TRANSFORMED;
	transformedMessage: TransformedChatMessage;
}

type ChatGenericMessage = {
  completion: ChatCompletionMessageParam | null;
  chatMessageType: CHAT_MESSAGE_TYPE;
  infoMessage?: string | null;
}

type ChatMessage = ChatUserMessageTransformed | ChatGenericMessage;
```

In fact, we could go even further and be more specific about _all_ our chat message types in the backend:
- ERROR_MSG and BOT_BLOCKED always have infoMessage but never completion
- SYSTEM always has completion (of type ChatCompletionSystemMessageParam) but never infoMessage
- BOT always has completion (of type ChatCompletionAssistantMessageParam) but never infoMessage. Note our current types are not quite up-to-date; interface "ChatCompletionMessage" which we are currently using in ChatGptReply is actually identical to "ChatCompletionAssistantMessageParam" so we should use the latter to match our other completion types)
- FUNCTION_CALL always has completion (of type ChatCompletionAssistantMessageParam or ChatCompletionToolMessageParam) but never infoMessage
- USER _either_ has infoMessage _or_ has completion (of type ChatCompletionUserMessageParam)

Note there are chat messages that come from the frontend, but they are only a subset of types (DEFENCE_ALERTED, DEFENCE_TRIGGERED, INFO, LEVEL_INFO, RESET_LEVEL), none have completion and all have infoMessage. So we might want another type to represent those front-end chat messages.

chriswilty (commented):
Some potential type cleanup, but this is good work 👍

backend/src/controller/chatController.ts
```diff
@@ -175,10 +171,12 @@ async function handleChatWithDefenceDetection(
 		defenceReport: combinedDefenceReport,
 		openAIErrorMessage: openAiReply.chatResponse.openAIErrorMessage,
 		reply: !combinedDefenceReport.isBlocked && botReply ? botReply : '',
-		transformedMessage: transformedMessage ?? undefined,
+		transformedMessage: messageTransformation?.transformedMessage ?? undefined,
 		wonLevel:
 			openAiReply.chatResponse.wonLevel && !combinedDefenceReport.isBlocked,
 		sentEmails: combinedDefenceReport.isBlocked ? [] : openAiReply.sentEmails,
+		transformedMessageInfo:
+			messageTransformation?.transformedMessageInfo ?? undefined,
```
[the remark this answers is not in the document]
pmarsh-scottlogic: I googled `null`s vs `undefined`s specifically for this while I was working on it. My conclusion was that they have different semantics: `null` say tells the programmer "this variable has no value, and we explicitly know that it is not meant to have a value, and we have assigned it as such", while `undefined` tells the programmer "this variable is yet to be assigned a value". I was happy enough with that distinction and so left in the slightly more verbose code. No worries to change it if you think it's easier to understand the other way.

backend/src/controller/chatController.ts
pmarsh-scottlogic: agreed on teams that it would be a separate PR for the same ticket.

pmarsh-scottlogic: Okay, now we have a problem where we need to check that a certain property in a chatMessage exists.

Here's a mini example:
Suppose we have two chat subtypes and a chat type that is the union of the former two.
```typescript
type ChatInfo = {
	info: string;
};

type ChatMessage = {
	message: string;
};

type Chat = ChatInfo | ChatMessage;
```

Now suppose we want to print the message from a given variable of type `chat`. We cannot do simply
```typescript
function printMessage(chat: Chat) {
  console.log(chat.message)
}
```

because `chatInfo` doesn't have the `message` property. Specifically, here's the ts error:

```
Property 'message' does not exist on type 'chat'.
  Property 'message' does not exist on type 'chatInfo'.ts(2339)
```

So instead we have to do something like.
```typescript
function printMessage(chat: Chat) {
  if ('message' in chat) {
    console.log(chat.message)
  }
}
```

pmarsh-scottlogic: Changes implemented here #803

backend/src/controller/chatController.ts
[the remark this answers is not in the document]
chriswilty: Well, you're breaking your own rule because you're setting transformedMessageInfo to undefined, rather than null.

Personally I don't have an issue with undefined in front-end code, because we are usually more interested in whether something is "nullish" - see optional chaining and nullish coalesce operators. Undefined is a little strange in javascript, because you can set a property or a "let" variable explicitly to undefined, hence, it doesn't mean "this variable is _yet_ to be assigned a value", it simply means "this variable doesn't currently have a value". Setting something to null to express intent is fine if you are then going to use that distinction in your code, but if not then it's extra cognitive load on the developer.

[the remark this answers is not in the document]
pmarsh-scottlogic: sweet, will make the change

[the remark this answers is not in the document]
pmarsh-scottlogic: done

backend/src/controller/chatController.ts
```diff
@@ -90,7 +90,7 @@ async function handleChatWithoutDefenceDetection(
 	chatHistory: ChatHistoryMessage[],
 	defences: Defence[]
 ): Promise<LevelHandlerResponse> {
-	const updatedChatHistory = createNewUserMessages(message, null).reduce(
+	const updatedChatHistory = createNewUserMessages(message, undefined).reduce(
```
chriswilty, line 93: You don't need to explicitly pass `undefined` here, just call the function with one arg.

backend/src/controller/chatController.ts
```diff
@@ -46,7 +46,7 @@ function combineChatDefenceReports(
 
 function createNewUserMessages(
 	message: string,
-	messageTransformation: MessageTransformation | null
+	messageTransformation: MessageTransformation | undefined
```
chriswilty, line 49: I'd advise using optional params if possible:

`messageTransformation?: MessageTransformation`

backend/src/defence.ts
```diff
@@ -256,18 +256,18 @@ function combineTransformedMessage(transformedMessage: TransformedChatMessage) {
...
 		? transformRandomSequenceEnclosure(message, defences)
 		: isDefenceActive(DEFENCE_ID.INSTRUCTION, defences)
 		? transformInstructionDefence(message, defences)
-		: null;
+		: undefined;
 
 	if (!transformedMessage) {
 		console.debug('No defences applied. Message unchanged.');
-		return null;
+		return undefined;
```
chriswilty, line 270: You should be able to use just `return;` unless the linter doesn't like that.

chriswilty (commented):
Nice, just some very minor advice on using undefined.

chriswilty (approved):
Beautiful 👌

PR #802, opened 2024-01-30
715 config update messages

dhinrichs-scottlogic:
## Description
Second try at this ticket. Started again from scratch. Previous PR can be found here: 
https://github.com/ScottLogic/prompt-injection/pull/778

Previously, when a user changed or reset the text or number inputs of the defence and model configurations, a message would pop up for 3 seconds to indicate to the user that the change has been made. This can easily be missed, which is why I moved this message to the chat window instead, where the user already gets updates when defences get toggled on/off.

Validation error messages remain underneath the input controls, and text/number inputs are now validated onChange rather than onBlur, giving users the chance to change their input. Error messages do not disappear after a time delay, but only disappear when the input is changed to a valid value. The error message field is announced to screen readers through aria-live="polite".

If a user keeps invalid input and commits it  (by pressing enter or changing focus), the message will reset on to the previously valid value. 

## Screenshots
image
image

## Notes
- changed two key/value pairs of the enum DEFENCE_ID: FILTER_USER_INPUT to INPUT_FILTERING and FILTER_BOT_OUTPUT to OUTPUT_FILTERING

### This is the second try to complete this ticket 
For documentation, I addressed the following issues mentioned in the [previous PR reviews](https://github.com/ScottLogic/prompt-injection/pull/778):
- kept validation where it existed, displaying an error message on the frontend onChange, 
- used addInfoMessage instead of addChatMessage
- moved validation into DefenceConfigurationInput (not DefenceConfiguration after all, based on discussions and a few tries)
- added error styling for the inputs and error message that persists beneath the input
- created separate function (addConfigUpdateToChat) to ensure no duplication when resetting and updating a config

## Concerns
- Due to the naming of the Information Defence, "defence" is repeated in the chat when it is activated, there is already https://github.com/ScottLogic/prompt-injection/issues/776 to address this. For the update info messages I did not include the word "defence" 
- There is currently no error messages for the user if the backend fails while we are changing a configuration. There's a https://github.com/ScottLogic/prompt-injection/issues/774to address this.
- The defence.name and the DEFENCE_ID are slightly different q/a LLM versus qa LLM. I couldn't change the enum to include the "/" in the enum key (which is what's being used in the function to update the config) and I didn't want to create a special conditional to change the output to the UI. Personally I think the difference between q/a and qa is acceptable for now. 

## Checklist

Have you done the following?

- [ ] Added tests


frontend/src/Theme.css
```diff
@@ -18,6 +18,7 @@
 	--main-text-accent-colour: #73d1d6;
 	--error-colour: #c41565;
 	--error-background-colour: #faebf2;
+	--error-colour-light: #ff75af;
```
chriswilty, line 21: I rather like this colour! I am assuming that our "standard" error colour is not contrasting enough on the dark background.

frontend/src/components/DefenceBox/DefenceConfigurationInput.tsx
```diff
@@ -34,40 +57,62 @@ function DefenceConfigurationInput({
...
+					onBlur={() => {
+						void setConfigurationValueIfDifferent(value);
+					}}
+					onKeyDown={inputKeyDown}
+					onKeyUp={inputKeyUp}
+					characterLimit={CONFIG_VALUE_CHARACTER_LIMIT}
+				/>
+				<p className="error-message" aria-live="polite">
+					{configValidated ? '' : 'Error: Invalid configuration value'}
+				</p>
```
chriswilty, line 92: You have this error JSX duplicated now. Better to declare your input separately, then return it in a fragment:

```
const inputElement = (inputType === 'text)
  ? (
    <ThemedTextArea ... />
  ) : (
    <ThemedNumberInput ... />
  );

return (
  <>
    {inputElement}
    {!configValidated && (
      <p className="error-message" aria-live="polite"> ... </p>
    )}
  </>
);
```

frontend/src/components/DefenceBox/DefenceMechanism.css
```diff
@@ -30,6 +31,7 @@
 
 .defence-mechanism-fieldset details[open] > summary {
 	border-style: inset;
+	background-color: var(--control-body-background-colour);
```
chriswilty, line 34: _sneaks tangential styles in without anyone noticing_ 🤭

frontend/src/components/ControlPanel/ControlPanel.css
```diff
@@ -35,3 +35,8 @@
 	border-bottom-right-radius: 0;
 	border-bottom-left-radius: 0;
 }
+
+.error-message {
```
chriswilty, line 39: "error-message" is a pretty generic name for a css class, so this style declaration could "leak" into other components that happen to use the same classname. Better to make the selector more specific:

```
.control-panel .control-collapsible-section .error-message {
```

frontend/src/components/ModelBox/ModelSelection.tsx
```diff
@@ -80,8 +87,8 @@ function ModelSelection({ chatModelOptions }: { chatModelOptions: string[] }) {
 
 				<div className="model-selection-info">
 					{errorChangingModel ? (
-						<p className="error">
-							Could not change model. You are still chatting to:
+						<p className="error-message" aria-live="polite">
```
chriswilty, line 90: While I understand you want to use the same styles in two different places (here and in DefenceConfigurationInput), it's usually better to have css defined alongside its component. This class is currently defined in ControlPanel.css.

I was going to suggest extracting a component for this error message, and defining the style there instead, however, that feels a bit overkill as I assume you only expect one style of error message within the whole Control Panel component hierarchy.

The other possibility would be to duplicate the css class and declare the styles in both places. To be honest, as it's only two places you're using it currently, and it's only two styles (one of which uses a theme property), I'm inclined to suggest that instead of tucking it away in ControlPanel.css. But then make the selectors more specific than just `.error-message`, as suggested elsewhere.

chriswilty (changes requested):
Looking nice! A few comments but not much to worry about.

frontend/src/Theme.css
dhinrichs-scottlogic: yep! I used https://www.learnui.design/tools/accessible-color-generator.html to find the closest match with good contrast

frontend/src/components/DefenceBox/DefenceMechanism.css
dhinrichs-scottlogic: 🤫

frontend/src/components/DefenceBox/DefenceConfigurationInput.tsx
```diff
@@ -2,28 +2,51 @@ import { KeyboardEvent, useEffect, useState } from 'react';
...
+	config,
 }: {
 	currentValue: string;
 	disabled: boolean;
 	inputType: 'text' | 'number';
-	setConfigurationValue: (value: string) => void;
+	setConfigurationValue: (value: string) => Promise<void>;
 	id: string;
+	defence: Defence;
+	config: DefenceConfigItem;
```
chriswilty, line 23: Not convinced you should be passing these two in, it gives DefenceConfigurationInput too much knowledge it shouldn't need.

You only need them so you can call validateDefence, but instead you could pass in a validate function prop that only takes the value to check, and then you would invoke validateDefence from the parent component where you already have access to defence and config.

frontend/src/components/DefenceBox/DefenceConfigurationInput.tsx
```diff
@@ -2,28 +2,51 @@ import { KeyboardEvent, useEffect, useState } from 'react';
...
+	const [configValidated, setConfigValidated] = useState<boolean>(true);
 
 	// update the text area value when reset/changed
 	useEffect(() => {
 		setValue(currentValue);
 	}, [currentValue]);
 
+	async function setConfigurationValueIfDifferent(value: string) {
+		const trimmedValue = value.trim();
+		if (trimmedValue !== config.value) {
```
chriswilty, line 36: config.value is the same as currentValue

frontend/src/components/ThemedInput/ThemedTextArea.tsx
```diff
@@ -92,11 +96,23 @@ function ThemedTextArea({
...
-		disabled,
-	});
+	// check config validated exists and is not false
+	const validInput = configValidated === false ? 'invalid-input' : '';
+
+	const textAreaClass = clsx(
+		'themed-input',
+		'themed-text-area',
+		spacing,
+		validInput,
```
chriswilty, line 111: clsx library is quite versatile; you can use this instead of needing to declare const validInput:

```
const textAreaClass = clsx(
  'themed-input',
  'themed-text-area',
  spacing,
  configValidated === false && 'invalid-input',
  { disabled }
);
```

Same idea in ThemedNumberInput.

chriswilty (changes requested):
One concern about props, but otherwise good to go 👍

frontend/src/components/ThemedInput/ThemedTextArea.tsx
chriswilty: Alternatively, you could "reverse polarity" and have the prop (and the state in the parent components) be "isInvalid".

It makes me a touch uncomfortable that you need to check that the value is specifically `false` for an optional prop, because if you don't pass in a prop value it'll be undefined, which is "falsy". Better to invert the prop so "falsy" and "undefined" can be treated the same.

Does that make sense?

frontend/src/components/DefenceBox/DefenceConfigurationInput.tsx
dhinrichs-scottlogic: I tried this, but then the onChange validation doesn't seem to work. 
Might be missing something, but let's look at it together tomorrow

frontend/src/components/DefenceBox/DefenceConfigurationInput.tsx
```diff
@@ -61,18 +56,16 @@ function DefenceConfigurationInput({
...
 			<ThemedTextArea
 				id={id}
 				content={value}
 				onContentChanged={setValue}
-				configValidated={configValidated}
-				validateInput={validateNewInput}
+				configInvalid={configInvalid}
+				validateInput={(value) => {
+					setConfigInvalid(!validateNewInput(value));
+				}}
```
chriswilty, line 68: As you have this function call in two places, it might be worth pulling it out into a function, as you originally had. It's also making my head spin a bit with the double negatives!

We might be able to improve the readability (or not!) by using a state var that is "positive" in this DefenceConfigurationInput component, i.e. "configIsValid", and only invert it where you pass it into the Themed components, e.g.

```
const [isConfigValid, setConfigValid] = useState<boolean>(true);

...

function validateConfigValue(value: string) {
  setConfigValid(validateNewInput(value));
}

...

<ThemedTextArea
  valueInvalid={!isConfigValid}
  validateInput={validateConfigValue}
```

Note that the Themed components are by design generic, so use "value" or "input" instead of "config" in prop names.

chriswilty (commented):
Looking good, just a suggestion which might or might not improve things! Try it and see?

frontend/src/components/DefenceBox/DefenceConfigurationInput.tsx
```diff
@@ -20,7 +20,7 @@ function DefenceConfigurationInput({
 }) {
 	const CONFIG_VALUE_CHARACTER_LIMIT = 5000;
 	const [value, setValue] = useState<string>(currentValue);
-	const [configInvalid, setConfigInvalid] = useState<boolean>(false);
+	const [inputInvalid, setInputInvalid] = useState<boolean>(false);
```
chriswilty, line 23: This was fine as "configInvalid" because this component is all about Configuration. It was just the Themed components where I was suggesting changing the prop name to "input" or "value".

dhinrichs-scottlogic: changed the state back to a positive declaration and passed the negation thereof to the inputs

frontend/src/components/DefenceBox/DefenceConfigurationInput.tsx
```diff
@@ -67,8 +66,8 @@ function DefenceConfigurationInput({
 				id={id}
 				content={value}
 				onContentChanged={setValue}
-				inputInvalid={inputInvalid}
-				validateInput={isInputInvalid}
+				valueInvalid={!isConfigValid}
+				validateInput={validateConfigValue}
```
chriswilty, line 70: Sorry Doro 😓 

I've just realised that we don't need to pass this` validateInput` function prop into the Themed components at all, because we already trigger onContentChanged whenever the value changes. So we can handle validation in this component instead, and just pass the "isInvalid" prop into the Themed components.

```
function handleChange(value: string) {
  setValue(value);
  setIsConfigValid(validateNewInput(value));
}

...

<ThemedTextArea
  content={value}
  contentInvalid={!isConfigValid}
  onContentChanged={handleChange}
```

dhinrichs-scottlogic: yep, makes sense, I made the changes

chriswilty: I'm going to make one annoying request... Sorry again!

As the other props use name "content" rather than "value", can we change the prop name for the two Themed components from "valueInvalid" to "contentInvalid", for consistency? Thanks 😊

chriswilty (approved):
👌 Beautiful, merge it!

Issue #744, opened 2024-01-05
Transformed blocked messages not showing correctly


gsproston-scottlogic:
# Bug report

## Description

If a user message is both transformed and blocked, then the transformed message will not show at all after a refresh or after switching levels.

## Reproduction steps

Steps to reproduce the behaviour:

1. Go to sandbox mode
2. Activate XML tagging
3. Activate character limit defence
4. Send a long chat message which will trigger the character limit defence
5. Wait for the reply
6. See that there's the original user message, the transformed user message, and the bot blocked message
7. Refresh the page OR switch levels
8. No transformed message

## Expected behaviour

The transformed message is still shown after the refresh or switching levels.

## Screenshots

Before refreshing or switching levels:

![image](https://github.com/ScottLogic/prompt-injection/assets/103250539/ab20e4e4-31b7-459a-9c0e-2d5897e06568)

After: 

![image](https://github.com/ScottLogic/prompt-injection/assets/103250539/5888f943-42a4-4d9b-ab42-399ca0911660)

## Software (please complete the following information):

- Browser: Edge

## Additional context

Could be related to #741 

## Acceptance criteria

**GIVEN** a message has been transformed (e.g. the user activates xml tagging and sends a message)
**AND** blocked
**WHEN** the page is refreshed
**THEN** the transformed message is still shown (below the original message)

**GIVEN** a message has been transformed (e.g. the user activates xml tagging and sends a message)
**AND** blocked
**WHEN** the user switches level and then back to the original level
**THEN** the transformed message is still shown (below the original message)

PR #803, opened 2024-01-30
741 fix backend chat history message types

pmarsh-scottlogic:
This PR is the result of a [change request](https://github.com/ScottLogic/prompt-injection/pull/800#discussion_r1467474623) in #800 . It was deemed a large enough change to warrant a second PR. This branch came of the branch for the first PR. This PR should be merged after #800.

## Description

Tightens up the types for the backend's chat message history. No functional change.

## Notes

- Renames `ChatHistoryMessage` to `ChatMessage`
- Introduces individual types for each of the `CHAT_MESSAGE_TYPE`s (enum). These specify which `CHAT_MESSAGE_TYPE`s need completions and which need info messages, and lets us be more specific about our completion types.
- `ChatMessage` is now a union of the above types, and all of them live in their own new file, chatMessage.ts

## Checklist

Have you done the following?

- [ ] Added tests


backend/src/models/chatMessage.ts
```diff
@@ -0,0 +1,92 @@
...
+	| ChatInfoMessage
+	| ChatUserMessage
+	| ChatDefenceTriggeredMessage
+	| ChatResetLevelMessage
+	| ChatDefenceAlertedMessage
+	| ChatBotMessage
+	| ChatLevelInfoMessage
+	| ChatSystemMessage;
+
+export type { ChatMessage, ChatSystemMessage };
```
pmarsh-scottlogic, line 92: annoying that I needed to export `ChatSystemMessage`. This was to do a cast on line 275 of openai.ts.

backend/src/openai.ts
```diff
@@ -271,8 +271,8 @@ async function chatGptChatCompletion(
 
 		// check to see if there's already a system role
 		const systemRole = chatHistory.find(
-			(message) => message.completion?.role === 'system'
-		);
+			(message) => message.chatMessageType === CHAT_MESSAGE_TYPE.SYSTEM
+		) as ChatSystemMessage | undefined;
```
pmarsh-scottlogic, line 275: We need this cast. Without it, the compiler does not infer that the `systemRole` will have a `.completion`, and there is an error when when we try to access it on line 284 (in the else statement a few lines down)

backend/test/integration/openai.test.ts
```diff
@@ -176,22 +175,32 @@ describe('OpenAI Integration Tests', () => {
...
+			// system role is added to the start of the chat history
+			expect(chatHistory[0].completion.role).toBe('system');
+			expect(chatHistory[0].completion.content).toBe(systemRoleDefault);
+
+			// rest of the chat history is in order
+			expect(chatHistory[1].completion.role).toBe('user');
+			expect(chatHistory[1].completion.content).toBe("I'm a user");
+			expect(chatHistory[2].completion.role).toBe('assistant');
+			expect(chatHistory[2].completion.content).toBe("I'm an assistant");
+		}
```
pmarsh-scottlogic, line 195: I've had to restructure a bunch of the tests to look like this, because we cannot assume that any chatMessage will have a completion. I've yet to think of a more concise way to express the expect statements.

chriswilty: Hmm. Can you not "coerce" into more specific type, e.g.

`const firstChatMessage = chatHistory[0] as ChatSystemMessage`?

backend/src/models/chatMessage.ts
chriswilty: This is fine, and you might need to export others based on another comment.

ChatMessage is a utility type so you don't need to use the full `ChatLevelInfoMessage | ChatSystemMessage | ChatUserMessage ...` type all over the place. It doesn't mean you shouldn't use the individual types if needed.

backend/src/utils/chat.ts
```diff
@@ -11,7 +11,10 @@ function pushMessageToHistory(
 	// remove the oldest message, not including system role message
 	// until the length of the chat history is less than maxChatHistoryLength
 	while (updatedChatHistory.length >= maxChatHistoryLength) {
-		if (updatedChatHistory[0].completion?.role !== 'system') {
+		if (
+			'completion' in updatedChatHistory[0] &&
+			updatedChatHistory[0].completion?.role !== 'system'
+		) {
 			updatedChatHistory.shift();
```
chriswilty, line 21: I believe we always expect the system role message at the very start of the chat history (or not at all). If so, it'd be more efficient to do something like this:

```
const spliceFrom = 
  'completion' in updatedChatHistory[0] && updatedChatHistory[0].completion?.role === 'system'
    ? 1
    : 0;

while (updatedChatHistory.length >= maxChatHistoryLength) {
  updatedChatHistory.splice(spliceFrom, 1);
}
```

Unsure if you can use `instanceof` instead of `in`, I have a feeling that only works with classes and interfaces:

```
const spliceFrom = updatedChatHistory[0] instanceof ChatSystemMessage ? 1 : 0;
```

backend/test/integration/openai.test.ts
pmarsh-scottlogic: I've found it's nicer to just compare the whole chatHistory in one expect statement

backend/test/integration/openai.test.ts
```diff
@@ -122,16 +118,19 @@ describe('OpenAI Integration Tests', () => {
 		// check the chat history has been updated
 		expect(chatHistory.length).toBe(1);
 		// system role is added to the start of the chat history
-		expect(chatHistory[0].completion?.role).toBe('system');
-		expect(chatHistory[0].completion?.content).toBe(systemRoleDefault);
+		expect('completion' in chatHistory[0]);
```
chriswilty: You appear to be missing an assertion here:

`expect( ... ).toBe(true)`

But this feels a little clunky. Can you not do this?

```
if ('completion` in chatHistory[0]) {
  ...
} else {
  throw Error(`unexpected chat message, type=${chatHistory[0].type}`);
}
```

backend/test/integration/openai.test.ts
```diff
@@ -176,22 +175,32 @@ describe('OpenAI Integration Tests', () => {
...
-		expect(chatHistory[2].completion?.role).toBe('assistant');
-		expect(chatHistory[2].completion?.content).toBe("I'm an assistant");
+		expect('completion' in chatHistory[0]);
+		expect('completion' in chatHistory[1]);
+		expect('completion' in chatHistory[2]);
+		if (
+			'completion' in chatHistory[0] &&
+			'completion' in chatHistory[1] &&
+			'completion' in chatHistory[2]
+		) {
```
chriswilty: As per previous comment, this seems a bit awkward. Also, I'd be tempted to test each message separately, in its own if...else block.

chriswilty (changes requested):
A few things I'm unsure of, but this is great work 🏆

backend/test/integration/openai.test.ts
pmarsh-scottlogic: no longer relevant, we now just do one big expect statement for the whole of the history

backend/test/integration/openai.test.ts
pmarsh-scottlogic: no longer relevant, this file no longer exists. Replaced with _unit/openai.test.ts_

backend/src/utils/chat.ts
pmarsh-scottlogic: I've refactored the whole lot to something that (I think) is more straightforward. Let me know what you think

backend/src/utils/chat.ts
```diff
@@ -12,19 +12,26 @@ function pushMessageToHistory(
...
+		chatHistory.length - maxChatHistoryLength + 1,
+		0
+	);
+	const updatedChatHistory = chatHistory.slice(
+		limitedHistoryStartIndex,
+		chatHistory.length
+	);
+
+	if (existingSystemRole) {
+		updatedChatHistory[0] = existingSystemRole;
```
chriswilty, line 31: Can't help thinking this is a touch more complex, not less. Within the scope of this function it's ok to be "mutating" our array copy to pop off messages, because all the logic is self-contained.

If you are intent on removing all the extra messages in one operation, rather than one at a time in a while loop, then how about this?

```
const updatedChatHistory = [...chatHistory, newMessage];
const messagesToRemove = updatedChatHistory.length - maxMessageLength;
if (messagesToRemove < 1) return updatedChatHistory;

const spliceFrom = updatedChatHistory[0].chatMessageType === CHAT_MESSAGE_TYPE.SYSTEM ? 1 : 0;
updatedChatHistory.splice(spliceFrom, messagesToRemove);
return updatedChatHistory;
```

chriswilty (commented):
Alternative suggestion!

backend/src/utils/chat.ts
pmarsh-scottlogic: tasty

backend/src/controller/chatController.ts
```diff
@@ -279,9 +276,8 @@ async function handleChatToGPT(req: OpenAiChatRequest, res: Response) {
 	if (updatedChatResponse.defenceReport.isBlocked) {
 		// chatReponse.reply is empty if blocked
 		updatedChatHistory = pushMessageToHistory(updatedChatHistory, {
-			completion: null,
 			chatMessageType: CHAT_MESSAGE_TYPE.BOT_BLOCKED,
-			infoMessage: updatedChatResponse.defenceReport.blockedReason,
+			infoMessage: updatedChatResponse.defenceReport.blockedReason ?? '',
```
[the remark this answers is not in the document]
chriswilty: Did you decide to do this, or leave it for another time? I've approved anyway 🎉

chriswilty (approved):
Sweet 👍

backend/src/controller/chatController.ts
```diff
@@ -373,10 +369,9 @@ function handleAddToChatHistory(req: OpenAiAddHistoryRequest, res: Response) {
 		req.session.levelState[level].chatHistory = pushMessageToHistory(
 			req.session.levelState[level].chatHistory,
 			{
-				completion: null,
 				chatMessageType,
 				infoMessage,
-			}
+			} as ChatMessage
```
pmarsh-scottlogic, line 374: Is this method type safe? Does it need to be? Is it problematic if, because of client error, a message has superfluous properties?

chriswilty: Hmm, good question. We might consider two improvements then:
- Create a type representing ChatMessages that originate from an API call (i.e. from the UI)
- Validate these messages on arrival, by checking their content conforms to our expected shape(s) and types.

Either or both of these could be done in a separate issue if you prefer, though you migth want to get them done now rather than kicking them down the road.

pmarsh-scottlogic: Adding extra types seems not useful if types don't exist at runtime. We'd need to add explicit checks that would work even when the project is compiled to js, non?

pmarsh-scottlogic: Okay, I've done my best shot at this, which may well could have been a separate PR's worth of work, so let me know if you want me to cherry pick commits across for ease of review.
- I've completely revamped the _models/chatMessage.ts_ file which defines the chat message types. All the chat message types which will (in the history) store the message in the `infoMessage` property are grouped together as `CHAT_MESSAGE_TYPE_AS_INFO`. This comes from an array of string literals. This same array is used to check that the message type in the request isn't bogus.
- While I was doing that I accidentally bumped into [this bug](https://github.com/ScottLogic/prompt-injection/issues/744) and solved it before realising there was already a ticket for it. I added an extra unit test to catch it.

this is a type-safe as I can get it, but it's kinda ugly and the names are confusing, so I'm happy to try another approach / take suggestions.

chriswilty (commented):
Will continue the review tomorrow. I'm probably better off looking at all the changes, instead of just the latest.

backend/src/models/chatMessage.ts
```diff
@@ -5,88 +5,73 @@ import {
...
+	'DEFENCE_TRIGGERED',
+	'LEVEL_INFO',
+	'RESET_LEVEL',
+	'ERROR_MSG',
+	'BOT_BLOCKED',
+	'USER',
+	'INFO',
+] as const;
+
+type CHAT_MESSAGE_TYPE_AS_INFO = (typeof chatMessageTypesAsInfo)[number];
```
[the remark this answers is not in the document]
pmarsh-scottlogic: Right. So as I was saying on teams, we have to things to distinguish.
- `"INFO"` which is a specific type of chat message which should be rendered in a certain way on the front end. This is distinct from, but the same kind of thing as `"LEVEL_INFO"` and `"LEVEL_RESET"` etc.
- chatMessages which store the the message in the `infoMessage` property. I was assigning these the type `chatMessageAsInfo`.

This was weird and the names were confusing. So what I've done is
- Rename "INFO" to "GENERIC_INFO". 
- Rename `chatMessageAsInfo` to `chatInfoMessage`.

And I've renamed the bits and bobs surrounding those too. The goal is to distinguish the two things. So now we have the idea of a chat info message, which is any message that appears in chat to give info, but doesn't get fed into the bot as a completion. Then we have the specific types that those messages can take, one of which is "GENERIC_INFO", and these types determine how they are rendered on the front end.

I think this is a satisfying resolution. What do you think?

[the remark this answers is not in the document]
chriswilty: Yep I think this will work fine, and you'll likely be revisiting anyway when you look at consolidating network requests.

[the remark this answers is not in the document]
chriswilty: "GENERIC_INFO"

Naming things is hard! How about "INFO_BUBBLE" or something like that? EVENT_INFO, INFORMATIVE, STATUS_UPDATE? Or just leave it as is cos you'll be revisiting this code anyway 😆

backend/src/models/chatMessage.ts
```diff
@@ -0,0 +1,70 @@
+import {
+	ChatCompletionAssistantMessageParam,
+	ChatCompletionMessageParam,
+	ChatCompletionSystemMessageParam,
+	ChatCompletionUserMessageParam,
+} from 'openai/resources/chat/completions';
+
+import { TransformedChatMessage } from './chat';
+
+const chatInfoMessageType = [
```
chriswilty, line 10: chatInfoMessageTypes (plural) ?

backend/src/controller/chatController.ts
```diff
@@ -162,11 +163,7 @@ async function handleChatWithDefenceDetection(
 
 	// if blocked, restore original chat history and add user message to chat history without completion
 	const updatedChatHistory = combinedDefenceReport.isBlocked
-		? pushMessageToHistory(chatHistory, {
-				completion: null,
-				chatMessageType: CHAT_MESSAGE_TYPE.USER,
-				infoMessage: message,
-		  })
+		? chatHistoryWithNewUserMessages
```
chriswilty, line 166: I've lost track of this PR a touch! Looks like you're keeping the transformed message now even if it was blocked, so I'm assuming that's a deliberate change.

backend/src/controller/chatController.ts
```diff
@@ -360,23 +356,26 @@ function handleGetChatHistory(req: OpenAiGetHistoryRequest, res: Response) {
...
 
-function handleAddToChatHistory(req: OpenAiAddHistoryRequest, res: Response) {
-	const infoMessage = req.body.message;
+function handleAddInfoToChatHistory(
+	req: OpenAiAddInfoToChatHistoryRequest,
+	res: Response
+) {
+	const infoMessage = req.body.infoMessage;
 	const chatMessageType = req.body.chatMessageType;
 	const level = req.body.level;
```
chriswilty, line 365: You could shorten these now if you wish:

```
const { chatMessageType, infoMessage, level } = req.body;
```

backend/src/controller/chatController.ts
```diff
@@ -360,23 +356,26 @@ function handleGetChatHistory(req: OpenAiGetHistoryRequest, res: Response) {
 	}
 }
 
-function handleAddToChatHistory(req: OpenAiAddHistoryRequest, res: Response) {
-	const infoMessage = req.body.message;
+function handleAddInfoToChatHistory(
+	req: OpenAiAddInfoToChatHistoryRequest,
+	res: Response
+) {
```
chriswilty, line 362: The more I think about this separate API call, the weirder it seems. In theory it allows us to inject an arbitrary message into the chat, of any chat message type. Ok we can't add a completion, but we could add a "BOT" message with an infoMessage property. Will the ChatGPT call simply ignore that message, as it has no completion property?

Anyhow, nothing to do here, just blurting out my thoughts as they form 😅

frontend/src/models/chat.ts
```diff
@@ -1,22 +1,19 @@
...
+	| 'GENERIC_INFO'
+	| 'USER'
+	| 'USER_TRANSFORMED'
+	| 'LEVEL_INFO'
+	| 'DEFENCE_ALERTED'
+	| 'DEFENCE_TRIGGERED'
+	| 'SYSTEM'
+	| 'FUNCTION_CALL'
+	| 'ERROR_MSG'
+	| 'RESET_LEVEL';
```
chriswilty, line 16: Enum begone! 🥳

chriswilty (commented):
All looks sane to me. Couple of questions / random thoughts, but essentially ready to approve once you're happy with it.

backend/src/controller/chatController.ts
pmarsh-scottlogic: well, not any more! They can only inject one that makes sense with the new checks I've added.
namely

```typescript
if (... && 
  chatInfoMessageType.includes(chatMessageType) &&
  ...
) {
  // do the stuff
}
```

backend/src/controller/chatController.ts
pmarsh-scottlogic: Ah, it was not so deliberate. I want the new messages in the history, but I do not want the completion there. I shall change that now

pmarsh-scottlogic: So now we need a version of `type ChatUserTransformedMessage` which doesn't include a completion. I was in two minds about whether I should just make the completion optional, or to make a whole new type that doesn't have the completion. I went with the former for lightweightness, but very happy to try the latter.

backend/src/controller/chatController.ts
```diff
@@ -62,23 +63,31 @@ function createNewUserMessages(
...
-			{
-				completion: {
-					role: 'user',
-					content: message,
-				},
-				chatMessageType: 'USER',
-			},
+			createAs === 'completion'
+				? {
+						completion: {
```
chriswilty, line 81: Is indentation squiffy here? One too many indents?

chriswilty (approved):
Yeah that looks nice 👌

backend/src/controller/chatController.ts
pmarsh-scottlogic: just Prettier doing its thing. Looks better in the ide

image

PR #804, opened 2024-01-31
596 fix problems with integrationlangchaintest

pmarsh-scottlogic:
## Description

Tidying up some code and making test a bit better

## Notes

- moves `vectorisedDocuments` and `initDocumentVectors` from _langchain.ts_ to _document.ts_.
- puts all document unit tests together into a folder.
  - moves tests from _integration/langchain.test.ts_ -> _unit/document.ts/initDocumentVectors.test.ts_.
  - renames/moves _unit/document.test.ts_ -> _unit/document.ts/getDocuments.test.ts_
- moves tests from _integration/langchain.test.ts_ -> _unit/langchain.ts/formatPromptEvaluation.test.ts_
- moves tests from _integration/langchain.test.ts_ -> _unit/langchain.ts/initialisePromptEvaluationModel.test.ts_
- tidies some test layouts and naming also.
- Individual tests are largely unchanged. Some mocks have been altered/added and some things reworded

## Checklist

Have you done the following?

- [ ] Added tests


backend/src/langchain.ts
```diff
@@ -195,17 +151,15 @@ async function queryPromptEvaluationModel(
 
 function formatEvaluationOutput(response: string) {
 	// remove all non-alphanumeric characters
-	try {
-		const cleanResponse = response.replace(/\W/g, '').toLowerCase();
+	const cleanResponse = response.replace(/\W/g, '').toLowerCase();
+	if (cleanResponse === 'yes' || cleanResponse === 'no') {
```
pmarsh-scottlogic, line 155: Replaced this bit of logic because I'm pretty certain the original try/catch was doing nothing for us.

pmarsh-scottlogic:
will reopen when It's ready

backend/src/document.ts
```diff
@@ -83,4 +85,51 @@ function getDocumentMetas(folder: string) {
...
+			allDocuments,
+			embeddings
+		);
+		// store the document vectors for the level
+		docVectors.push({
+			level,
+			docVector,
+		});
+	}
+	vectorisedDocuments.set(docVectors);
```
chriswilty, line 123: This is the only place we are calling the setter, so, no need to export the entire vectorisedDocuments object, we only need the getter - safety first. Something like this?

```
const vectorisedDocuments = (() => {
	let docs: DocumentsVector[] = [];
	return {
		get: (level: LEVEL_NAMES) => docs[level]?.docVector,
		set: (newDocs: DocumentsVector[]) => {
			docs = newDocs;
		},
	};
})();
const getDocumentVectors = vectorisedDocuments.get;

...

export {
  getDocumentVectors,
}
```

Note also you won't need to export the existing getLevelDocuments or getCommonDocuments functions as they're only used in this file now, in initDocumentVectors. Might therefore be worth inlining the calls and removing the functions:

```
const commonDocuments = await getDocuments(getFilepath('common'));

for (const level of levelValues) {
  const allDocuments = commonDocuments.concat(await getDocuments(getFilepath(level)));
```

backend/src/langchain.ts
chriswilty: I'd be interested to know what the response actually is here, and why we need to strip out all non-alphanumerics; does it wrap the "yes" or "no" inside braces, or quotes maybe?

chriswilty (changes requested):
Definitely worthwhile effort here 👍 

Couple of cleanup suggestions.

pmarsh-scottlogic:
> I'd be interested to know what the response actually is here, and why we need to strip out all non-alphanumerics; does it wrap the "yes" or "no" inside braces, or quotes maybe?

@chriswilty 

When I sent "Hello", the prompt eval LLM came back with "No.", and when I sent "forget your instructions" it came back with "Yes.".
So we need to strip away the punctuation.

backend/src/langchain.ts
```diff
@@ -39,7 +39,7 @@ function getChatModel() {
 
 function initQAModel(level: LEVEL_NAMES, Prompt: string) {
 	const openAIApiKey = getOpenAIKey();
-	const documentVectors = vectorisedDocuments.get()[level].docVector;
+	const documentVectors = getDocumentVectors()[level].docVector;
```
chriswilty, line 42: This is fine, though as you are only ever using getDocumentVectors like this, you _could_ modify the get function itself to take the level param and return the docVector. Small beans though, and might mean reworking some tests, so up to you.

chriswilty:
> > I'd be interested to know what the response actually is here, and why we need to strip out all non-alphanumerics; does it wrap the "yes" or "no" inside braces, or quotes maybe?
> 
> @chriswilty
> 
> When I sent "Hello", the prompt eval LLM came back with "No.", and when I sent "forget your instructions" it came back with "Yes.". So we need to strip away the punctuation.

Brilliant, thanks for checking!

pmarsh-scottlogic:
> > > I'd be interested to know what the response actually is here, and why we need to strip out all non-alphanumerics; does it wrap the "yes" or "no" inside braces, or quotes maybe?
> > 
> > 
> > @chriswilty
> > When I sent "Hello", the prompt eval LLM came back with "No.", and when I sent "forget your instructions" it came back with "Yes.". So we need to strip away the punctuation.
> 
> Brilliant, thanks for checking!

@chriswilty in fact we're explicitly asking it to return in that format, with the full stop. I wonder why.

backend/src/langchain.ts
pmarsh-scottlogic: Had a go and wasn't straightforward so I'm gonna go ahead and ignore it

backend/src/document.ts
pmarsh-scottlogic: @chriswilty 

Because we removed getLevelDocuments and getCommonDocuments, I've deleted the unit tests for those. But now we're left with `initDocumentVectors` being pretty untestable, because it's function is to set the documentVectors variable. So I guess that we can test that it's set correctly. But its type is from langchain and will probably be pretty horrible to test. Any suggestions? Leave it be?

backend/test/unit/document.ts/initDocumentVectors.test.ts
```diff
@@ -0,0 +1,38 @@
...
+		}),
+	};
+});
+
+test('GIVEN application WHEN application starts THEN document vectors are loaded for all levels', async () => {
+	mockSplitDocuments.mockResolvedValue([]);
+
+	await initDocumentVectors();
+
+	const numberOfCalls = 4 + 1; // number of levels + common
```
chriswilty, line 35: 4 is a magic number; can we not derive it from the enum?

backend/src/document.ts
pmarsh-scottlogic: Seems pretty low priority

chriswilty:
@pmarsh-scottlogic I'm happy to approve this anyway. If you need any help with merge conflicts, just ask (tomorrow!)

chriswilty (approved):
Lovely 💎

PR #811, opened 2024-02-01
810: Session vs non-session routes

### LINKED ISSUE #810: Split Express routing into session vs non-session
In order to prevent unnecessary sessions being created when we hit e.g. the health check endpoint (which the load balancer does regularly) we neede to re-configure the express middleware and routes so that the session is only used on endpoints that need to read or write session data.

This work was originally done on the cloud infrastructure branch, but I'd like it in dev first, so that the cloud branch _only_ has infra code.

chriswilty:
## Description

We need to split API routes into session and non-session endpoints, so that we are not generating or noop-updating the session when we don't need to. This is particularly important for cloud deployment, where our load-balancer hits our healthcheck every 30secs, creating a new session every time 😱 

Resolves #810 

## Checklist

Have you done the following?

- [ ] Added tests


backend/src/app.ts
```diff
@@ -1,74 +1,16 @@
...
-import { importMetaUrl } from './importMetaUtils';
-import { ChatModel, defaultChatModel } from './models/chat';
-import { LevelState, getInitialLevelStates } from './models/level';
-import { router } from './router';
+import nonSessionRoutes from './nonSessionRoutes';
+import sessionRoutes from './sessionRoutes';
 
-dotenv.config();
-
-declare module 'express-session' {
```
chriswilty, line 15: Much of this file has been moved to sessionRoutes.ts, the rest in nonSessionRoutes.ts

backend/src/nonSessionRoutes.ts
```diff
@@ -0,0 +1,21 @@
...
+
+router.use(
+	'/documents',
+	express.static(
+		fileURLToPath(new URL('../resources/documents', importMetaUrl()))
+	)
+);
+router.get('/documents', handleGetDocuments);
+router.get('/health', handleHealthCheck);
+router.get('/systemRoles', handleGetSystemRoles);
```
chriswilty, line 19: There might be one or two more endpoints that don't need a session, e.g. validModels, but these will do for now.

backend/package.json
```diff
@@ -23,8 +23,7 @@
 		"memorystore": "^1.6.7",
 		"openai": "^4.19.0",
 		"openai-chat-tokens": "^0.2.8",
-		"pdf-parse": "^1.1.1",
-		"react": "^18.2.0"
```
pmarsh-scottlogic, line 27: good spot lol

backend/src/langchain.ts
```diff
@@ -20,11 +20,12 @@ import {
...
 // store vectorised documents for each level as array
 const vectorisedDocuments = (() => {
-	let docs: DocumentsVector[] = [];
+	const docs: DocumentsVector[] = [];
 	return {
 		get: () => docs,
 		set: (newDocs: DocumentsVector[]) => {
-			docs = newDocs;
+			while (docs.length > 0) docs.pop();
+			docs.push(...newDocs);
```
pmarsh-scottlogic, line 28: Is there a particular reason why you did this? Just curious, not upset with it.

backend/src/models/chat.ts
```diff
@@ -141,20 +141,15 @@ export type {
 	ChatDefenceReport,
 	ChatGptReply,
 	ChatMalicious,
+	ChatModel,
+	ChatModelConfiguration,
```
pmarsh-scottlogic, line 145: another good spot 👍

backend/src/sessionRoutes.ts
```diff
@@ -0,0 +1,133 @@
...
+if (!sessionSigningSecret) {
+	console.error(
+		'SESSION_SECRET not found in environment vars, cannot continue!'
+	);
+	process.exit(1);
+}
+
+const router = express.Router();
+
+const stage = process.env.NODE_ENV;
```
pmarsh-scottlogic, line 51: stage seems like a weird name, but then I'm not familiar with the terminology.

pmarsh-scottlogic (changes requested):
Struggling to run this locally. Some (but not all) endpoints come back with CORS errors

loading up the frontend

image

Sending a chat message

image

testing on postman the endpoints seem fine.

backend/src/langchain.ts
chriswilty: There was - I figured probably better to use a const within the closure; we can then be certain we are not holding onto someone else's copy of the document array, and therefore our copy will not change without us knowing about it.

With the original code, we were replacing our reference entirely, so fully at the mercy of any other code with access to that reference. Of course, the contents are only shallow-copied so the actual items could be mutated, but hopefully we are avoiding as much mutation as possible now in our codebase.

backend/src/sessionRoutes.ts
chriswilty: Yeah it's an AWS thing 🤐 

They use "environment" to mean "region and account", whereas what we know as environments are called stages. Which is doubly odd when it's quite common to have a "staging" environment, so then you'd have "staging" stage 😕

chriswilty:
@pmarsh-scottlogic Ah yes sorry, you'll need to replace your backend `.env` with the new `.env.example` as I added the cors origin into there.

backend/src/langchain.ts
pmarsh-scottlogic: right, makes sense

backend/src/sessionRoutes.ts
pmarsh-scottlogic: Ah well! Can't win them all

pmarsh-scottlogic (approved):
Works a charm

backend/test/integration/langchain.test.ts
```diff
@@ -49,16 +51,15 @@ jest.mock('langchain/embeddings/openai', () => {
 	};
 });
 
-class MockMemoryVectorStore {
-	asRetriever() {
-		mockAsRetriever();
-	}
-}
```
chriswilty, line 56: @pmarsh-scottlogic It seemed unnecessary to create a class, so I inlined it below.

backend/src/document.ts
```diff
@@ -28,6 +28,7 @@ async function getDocuments(filePath: string) {
 		'.csv': (path: string) => new CSVLoader(path),
 	});
 	const docs = await loader.load();
+	console.debug(`${docs.length} documents found`);
```
chriswilty, line 31: This is the line that exposed we weren't stubbing the loader to return anything. So I simply added that in my test changes.

backend/test/integration/langchain.test.ts
```diff
@@ -115,28 +116,28 @@ jest.mock('@src/openai', () => {
...
-	mockRetrievalQAChain.call.mockRestore();
-	mockFromLLM.mockRestore();
-	mockFromTemplate.mockRestore();
-});
+	afterEach(() => {
+		mockPromptEvalChain.call.mockReset();
+		mockRetrievalQAChain.call.mockReset();
+		mockFromLLM.mockReset();
+		mockFromTemplate.mockReset();
+		mockLoader.mockReset();
```
pmarsh-scottlogic, line 135: could you use `mockClear()` instead and then not have to do `mockLoader.mockResolvedValue([]);` in the beforeEach?

[doc](mockLoader.mockResolvedValue([]);)

chriswilty: Done

pmarsh-scottlogic (approved):
lovely

PR #817, opened 2024-02-05
CDK API stack with Fargate / ALB

### LINKED ISSUE #385: Fargate/ECS stack for back-end deployment
Use Typescript-CDK to define a stack for deploying the back-end container to AWS. To begin with, we will deploy the stack manually.

CDK example:
https://docs.aws.amazon.com/cdk/v2/guide/ecs_example.html

Node container and other Fargate/ECS/ELB considerations:
https://medium.com/@arliber/aws-fargate-from-start-to-finish-for-a-nodejs-app-9a0e5fbf6361

**Notes**
- #379 must be completed first!
- What to do about CORS before our UI stack is deployed? Do we allow all, or allow localhost:5173 (vite's devserver default port)?
- Until we assess and tighten the security / roles of the deployed resources, we should remove (destroy) the deployed stage every time after testing. This is trivial with CDK.

**Questions**
- Do we want to start with a shared, hard-coded API key (injected into the environment at stack synthesis) or adopt secrets manager from the start?
- Do we want to set strict rate limits from the start, for safety?
- Do we want at least two instances to begin with, to verify load balancing?

**Out of Scope**
- UI stack will be defined in a separate issue
- Pipeline deployment of both stacks will be covered in a subsequent issue

**Acceptance Criteria**
CDK code for (manually) deploying a VPC / ELB / Fargate / ECS stack.
Deployed, verified then destroyed using CDK CLI.

chriswilty:
## Description

Fargate-managed resources defined as CDK constructs, written in typescript.
https://docs.aws.amazon.com/cdk/v2/guide/getting_started.html

Part of #385 

## Notes

- CDK generates some stuff for us when we run `cdk init`, such as a cdk.json config file. All you need to know about that file right now is, it includes a heap of settings to ensure we have sensible protections and defaults on our deployed resources.
- Our entire "application" is defined in `bin/cloud.ts` and it uses "stacks" to define its resources. Although it is common to have just one, you can actually define as many different applications as you wish, e.g. linux vs windows stacks, or maybe you want to introduce a feature in an experimental stack that won't interfere with your stable deployments. By convention these all sit in the bin/ directory.
- All resource stacks sit in the lib/ directory. A stack is simply a collection of resources deployed to AWS together; stacks tend to contain related resources, but more generally they are a way to organise your resources into groups. CDK works out any dependencies between stacks so that they are deployed in the correct order. For example, in a later PR I will add the UI stack, which defines a URL on which the UI is available; we need to tell the API stack that URL, so it can be added to _allowed origins_ in CORS configuration. This means that the UI stack will be deployed first, followed by the API stack, regardless of the order they are defined in `bin/cloud.ts`.
- Our API stack is defined in `lib/api-stack.ts`. Observe that we define a docker Image, a VPC (virtual private cloud) with all its networking resources, and a Fargate service with cluster and load-balancer, all in a handful of lines of code; this produces over a thousand lines of CloudFormation template code! We also define an integration to AWS Secrets Manager to access API key and secret, and we pass those values to our docker container as env vars, so that they are available to our express server on startup. That happens within AWS at deployment time, so we don't need to store those values locally, nor send them over the network.
- Finally, note we are making use of our server healthcheck endpoint. This is how we tell Fargate if the server within each container is up and running, and Fargate will automatically kill and recreate nodes that are unhealthy, up to our desired minimum number of nodes (in our case, just one). @pmarsh-scottlogic This is why I wanted to remove the session cookie from that endpoint, because Fargate will ping the server regularly (every 30secs by default), and if every time we initiated a new session, that would be one of them dreaded resource leaks 😰 

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/cloud/.npmignore
+++ b/cloud/.npmignore
@@ -1,6 +1,0 @@
-*.ts
-!*.d.ts
-
-# CDK asset staging directory
-.cdk.staging
-cdk.out
--- a/cloud/.npmignore
+++ b/cloud/.npmignore
@@ -1,0 +1,6 @@
+*.ts
+!*.d.ts
+
+# CDK asset staging directory
+.cdk.staging
+cdk.out
--- a/cloud/.prettierignore
+++ b/cloud/.prettierignore
@@ -1,0 +1,2 @@
+node_modules
+cdk.out
--- a/cloud/.prettierignore
+++ b/cloud/.prettierignore
@@ -1,2 +1,1 @@
-node_modules
 cdk.out
--- a/cloud/README.md
+++ b/cloud/README.md
@@ -1,0 +1,17 @@
+# CDK project for Prompt Injection
+
+This project uses AWS CDK (using TypeScript) to define CloudFormation templates for remote deployment of all resources
+for the Prompt Injection application.
+
+The architecture is a typical containerized node-express API managed by AWS Fargate and ECS, with load-balancing (ELB),
+plus an S3-hosted UI served through CloudFront, all secured using Cognito.
+
+The `cdk.json` file tells the CDK Toolkit how to execute your app.
+
+## Commands
+
+- `npm run cdk:synth` - generates the CloudFormation templates, outputting into `./cdk.out` dir.
+- `npm run cdk:deploy` - deploys the application stacks into the remote DEV stage.  
+  _Caution! This will overwrite existing resources, so ensure the team are notified before running this manually._
+
+Note that once the CDK Pipeline is in place, DEV deployment will happen automatically on merging into dev branch.
--- a/cloud/bin/cloud.ts
+++ b/cloud/bin/cloud.ts
@@ -1,0 +1,27 @@
+#!/usr/bin/env node
+import { App } from 'aws-cdk-lib';
+import 'source-map-support/register';
+
+import {
+	appName,
+	resourceDescription,
+	stackName,
+	stageName,
+	ApiStack,
+} from '../lib';
+
+const app = new App();
+const tags = {
+	Project: appName,
+	App: appName,
+	Environment: stageName(app).toUpperCase(),
+	IaC: 'CDK',
+};
+
+const generateStackName = stackName(app);
+const generateDescription = resourceDescription(app);
+
+new ApiStack(app, generateStackName('api'), {
+	tags,
+	description: generateDescription('API stack'),
+});
--- a/cloud/bin/cloud.ts
+++ b/cloud/bin/cloud.ts
@@ -4,17 +4,18 @@ import 'source-map-support/register';
 
 import {
 	appName,
+	environmentName,
 	resourceDescription,
 	stackName,
-	stageName,
 	ApiStack,
 } from '../lib';
 
 const app = new App();
 const tags = {
-	Project: appName,
-	App: appName,
-	Environment: stageName(app).toUpperCase(),
+	owner: appName,
+	classification: 'unrestricted',
+	'environment-type': environmentName(app),
+	'keep-alive': '8-6-without-weekends',
 	IaC: 'CDK',
 };
 
--- a/cloud/cdk-developer-policy.yaml
+++ b/cloud/cdk-developer-policy.yaml
@@ -1,0 +1,52 @@
+Resources:
+  PermissionsBoundary:
+    Type: AWS::IAM::ManagedPolicy
+    Properties:
+      PolicyDocument:
+        Statement:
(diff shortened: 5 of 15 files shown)


cloud/lib/index.ts
pmarsh-scottlogic, line 1: What's the deal with having bin/cloud.ts and having everything else in the lib/ directory? Is that just convention?

cloud/.prettierignore
```diff
@@ -0,0 +1,2 @@
+node_modules
+cdk.out
```
pmarsh-scottlogic, line 2: github's wanting a new line at the end here

cloud/README.md
```diff
@@ -0,0 +1,17 @@
...
+This project uses AWS CDK (using TypeScript) to define CloudFormation templates for remote deployment of all resources
+for the Prompt Injection application.
+
+The architecture is a typical containerized node-express API managed by AWS Fargate and ECS, with load-balancing (ELB),
+plus an S3-hosted UI served through CloudFront, all secured using Cognito.
+
+The `cdk.json` file tells the CDK Toolkit how to execute your app.
+
+## Commands
+
```
pmarsh-scottlogic, line 12: worth adding a note about signing into aws account first? Or will any cloud person find this obvious?

cloud/bin/cloud.ts
```diff
@@ -0,0 +1,27 @@
...
+const app = new App();
+const tags = {
+	Project: appName,
+	App: appName,
+	Environment: stageName(app).toUpperCase(),
+	IaC: 'CDK',
+};
+
+const generateStackName = stackName(app);
+const generateDescription = resourceDescription(app);
```
pmarsh-scottlogic, line 22: Ah, so I'm guessing you're gonna use these again for the UI stack

cloud/bin/cloud.ts
```diff
@@ -0,0 +1,27 @@
...
+import {
+	appName,
+	resourceDescription,
+	stackName,
+	stageName,
+	ApiStack,
+} from '../lib';
+
+const app = new App();
+const tags = {
```
pmarsh-scottlogic, line 14: Worth adding an owner tag? I remember being asked for that on grad project, but don't know what the company policy is

cloud/cdk-developer-policy.yaml
pmarsh-scottlogic, line 1: not gonna bother reviewing this one

cloud/.npmignore
pmarsh-scottlogic, line 1: As I understand it, we're asking npm to not include these things in the package. But when will we ever package up the project? Is this just a default file? Why would we exclude all of the ts files?

cloud/lib/resourceNamingUtils.ts
```diff
@@ -0,0 +1,15 @@
...
+export const appName = 'SpyLogic';
+
+export const stageName = (construct: Construct) =>
+	(construct.node.tryGetContext('STAGE') as string) || 'dev';
+
+export const resourceName = (construct: Construct) => (suffix: string) =>
+	`${stageName(construct)}-${appName}-${suffix}`.toLowerCase();
+
+export const resourceDescription = (construct: Construct) => (prefix: string) =>
+	`${prefix} for ${appName} (${stageName(construct)})`;
```
pmarsh-scottlogic, line 12: Seems you were hungry for some Curry when you were writing this.

Any particular reason for things to be defined this way? Convention? Speed?

pmarsh-scottlogic (commented):
Got a bunch of questions. Nothing stopping me from approving though

pmarsh-scottlogic:
kudos for lovely description by the way

cloud/lib/index.ts
chriswilty: Exactly that! Cloud apps are in bin and stacks / stages in lib.

cloud/README.md
chriswilty: Well, it depends how the user is going to authenticate with AWS. They may still be using the long-lived token way of authenticating (even though that is discouraged) and in that case they'd not need to login. As this code is currently planned to make its way into the dev branch, and therefore be part of the open-source offering, I'm inclined to let the user manage that side of things.

pmarsh-scottlogic (approved):

cloud/bin/cloud.ts
chriswilty: @pmarsh-scottlogic FYI We need to request a new, project-specific AWS account instead of using Dev Edinburgh. Then I think tags will be less important. I've requested an account today.

Issue #808, opened 2024-02-01
Connect cognito userpool with Application Load Balancer


chriswilty:
Here is a [SO post on integrating cognito with Application Load Balancer (ALB)](https://stackoverflow.com/q/71124324). It is straightforward, but requires the ALB to be secure, hence this issue is **blocked by #806**.

Also dependent on #807

**Update - 02/04/2024**

This turned out to be a heap more difficult than expected. ALB does not provide any way to verify an auth token in a request header or cookie, rather, ALB wants to _provide_ the connection to your Identity Provider. This is in contrast to API Gateway which provides an explicit way to verify a provided auth token.

In theory we could allow ALB to provide the auth connector, however, that would mean using the Cognito hosted UI, which does not provide enough flexibility to allow matching our SpyLogic style. In contrast, AWS Amplify provides its own UI with greater ability to customise, plus, if that wasn't good enough we could provide our own UI and hook it up to the auth endpoints provided by Cognito. So, we have no choice but to initiate auth through the UI and simply verify the token at the load balancer...

... which is not possible! There is, however, another potential solution. We could use a CloudFront distribution as a (non-caching) proxy to the ALB with an Edge Lambda function to verify the token, and if it is valid, add a custom header to the request, containing a UUID generated during stack deployment which the ALB will therefore know about. The ALB can be configured with a simple rule to forward traffic to the target group (i.e. fargate service) only if it contains this header with the expected UUID value, and the default rule can be amended to return 401 Unauthorized. As this traffic is entirely within the AWS network, the token value should never leak to the outside  world. The end result is that traffic _not_ routed through cloudfront will always return 401, and traffic that _is_ routed through cloudfront must have a valid token - which Amplify will take care of in the UI.

Issue #807, opened 2024-02-01
Cognito Authentication for UI


chriswilty:
We need a cognito userpool integrated with our ScottLogic azure AD. We will also want to add users directly to the userpool, for client / guest access, hence we will not use Azure AD directly for auth.

Can use project-overwatch repo for an example of Cognito userpools via CDK.

Check if any other projects have used cognito in this way, directly linking to main Azure AD, instead of using a sandbox. If not, there should be instructions on our SL intranet for AD integration.

Dependent on #388.

PR #818, opened 2024-02-05
807: CDK auth stack (AWS Cognito)

chriswilty:
## Description

Cognito-based CDK auth stack.

Main parts of our Cognito resources are
- `UserPool` - just as it sounds, defines a pool of users allowed to authenticate with our resources
- `IdentityProvider` - provides an integration with Azure AD, so that all ScottLogic users can gain access to the application
- `UserPoolClient` - determines how the userpool will be accessed, including the auth flow to use, any identity provider(s) to integrate with, and other stuff such as token lifetimes

Prerequisite for #807 and #808

## Concerns

I am adding this stack to the CDK app, even though it contains some commented-out code and the stack isn't even being deployed currently, because it was looking a touch tricky to extract it. I will shortly come to the auth stuff anyway, once the Route 53 domain-based stack is added.

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/cloud/bin/cloud.ts
+++ b/cloud/bin/cloud.ts
@@ -9,6 +9,7 @@ import {
 	stackName,
 	ApiStack,
 } from '../lib';
+import { AuthStack } from '../lib/auth-stack';
 
 const app = new App();
 const tags = {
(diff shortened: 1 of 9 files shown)


cloud/bin/cloud.ts
```diff
@@ -21,7 +21,18 @@ const tags = {
 const generateStackName = stackName(app);
 const generateDescription = resourceDescription(app);
 
+// Don't need this stack, yet... Or ever? Will ask Pete C.
```
chriswilty, line 24: Since confirmed, yes we will need it eventually!

pmarsh-scottlogic (approved):

PR #820, opened 2024-02-06
Import services methods in object rather than separately

pmarsh-scottlogic:
## Description

In the frontend we now import service methods like `import * as someService from '@src/service/someService'`
then do `someService.serviceMethod()` in the code.

This is to clear up the many instances of confusing code like
```typescript
function addInfoMessage(message: string) {
    addChatMessage({
        message,
        type: 'GENERIC_INFO',
    });
    // asynchronously add message to chat history
    void addInfoMessageToChatHistory(message, 'GENERIC_INFO', currentLevel);
}
```

where it is not clear what things will call the api and which things are only doing things in the frontend.

## Concerns

- I imagine this comes at a minor performance cost, but I think the readability is worth it
- I wonder if there is a way to enforce that we import from our services like this, so that we are sure we will do it consistently in the future.

## Checklist

Have you done the following?

- [ ] Linked the relevant Issue (no issue)
- [ ] Added tests (no functional change. Not worth testing)


chriswilty:
> I wonder if there is a way to enforce that we import from our services like this

There is, kind of: use an index.t file in @src/service, and inside that put e.g.

```
import * as chatService from './chatService';
import * as configService from './configService';

export {
  chatService,
  configService,
}
```

Then you would import your services from the directory:

```
import { chatService } from '@src/service';
```

You _can_ enforce directory-level imports in eslint config, if I recall, but we're not ready for that yet as we'd need index files throughout our codebase.

frontend/src/components/MainComponent/MainComponent.tsx
```diff
@@ -7,19 +7,10 @@ import { CHAT_MESSAGE_TYPE, ChatMessage } from '@src/models/chat';
...
-	configureDefence,
-	getDefences,
-	resetDefenceConfig,
-} from '@src/service/defenceService';
-import { clearEmails, getSentEmails } from '@src/service/emailService';
-import { healthCheck } from '@src/service/healthService';
+import * as chatService from '@src/service/chatService';
+import * as defenceService from '@src/service/defenceService';
+import * as emailService from '@src/service/emailService';
+import * as healthService from '@src/service/healthService';
```
chriswilty, line 13: Health Service 🚑

frontend/src/components/MainComponent/MainComponent.tsx
```diff
@@ -64,7 +55,7 @@ function MainComponent({
 	// called on mount
 	useEffect(() => {
 		// perform backend health check
-		healthCheck().catch(() => {
+		healthService.healthCheck().catch(() => {
```
chriswilty, line 58: This is interesting. Once you consolidate the service calls, I imagine you won't need this health check, because the MainComponent will mount at the same time as the "fetch level data" request is fired off.

chriswilty (commented):
👍 Happy to approve, but thought you might want to use index exports/imports as suggested.

frontend/src/components/MainComponent/MainComponent.tsx
pmarsh-scottlogic: yet we need the health endpoint for the AWS load balancer, right?

chriswilty: Correct. Seems we'd not need to call it from the frontend, but we'll still want the endpoint defined in the backend, for the load balancer.

chriswilty (approved):
Beautiful 👌

PR #829, opened 2024-02-07
782 give every modal a header with title and close button

dhinrichs-scottlogic:
# Description
Every Modal now has a heading and a close button with the text "close" and an x icon. 

## Screenshots
image
image
image

## Notes
- new OverlayHeader component
- new x icon svg 
- style update for DocViewer and Handbook

## Concerns
- I took the svg from [here](https://uxwing.com/red-x-line-icon/) and then removed anything <s> I didn't understand </s> that seemed unnecessary, so would be great to double check that I didn't remove anything important. 

## Checklist

Have you done the following?

- [x] Added tests


frontend/src/components/Overlay/OverlayHeader.tsx
```diff
@@ -0,0 +1,30 @@
...
+	heading: string;
+	iconColor?: string;
+}
+
+function OverlayHeader({
+	closeOverlay,
+	heading,
+	iconColor,
+}: OverlayHeaderProps) {
+	const iconFill: string = iconColor ?? '#000';
```
pmarsh-scottlogic, line 16:
```suggestion
	const iconFill: = iconColor ?? '#000';
```

the ts compiler will infer the type from the expression

pmarsh-scottlogic (commented):
I've not done a proper review yet, but I noticed that on the handbook the header is overlapping the top of the scroll bar:

image

Then I realised how the handbook header was structured: 
- The `overlay-header` div sits at the top with `position:fixed` while the `handbook-overlay` div sits tucked underneath, and it's up to some clever margins to nudge the content and spine down a bit below the header.

image

image

And while this looks grand if you get the margins _just so_, It seems vulnerable to visual bugs down the line.

Ideally we'd have the `overlay-header` actually take up space in the div, and have the spine and content fill up the remaining space. I had a go at that in [this separate branch](https://github.com/ScottLogic/prompt-injection/tree/handbook-overlay-suggestion) (You can click compare and pull request to see the changes). It would require some more css fiddling to get the other modals looking good again though, since they all depend on the overlay-header having `position:fixed`.

frontend/src/components/DocumentViewer/DocumentViewBoxNav.tsx
```diff
@@ -1,29 +1,28 @@
...
 	onNext,
-}: DocumentViewBoxHeaderProps) {
+}: DocumentViewBoxNavProps) {
 	const documentNumber = `${documentIndex + 1} out of ${numberOfDocuments}`;
 
 	return (
-		<div className="view-documents-header">
-			<h2>View documents</h2>
+		<div className="view-documents-nav">
 			<div className="nav">
```
chriswilty, line 24: Turns out you don't need this extra div, you can simply have your nav directly within the "view-documents-nav".

frontend/src/components/DocumentViewer/DocumentViewBoxNav.tsx
```diff
@@ -1,29 +1,28 @@
...
-				<div className="button-container">
+				<nav className="button-container">
```
chriswilty, line 25: Does this need an aria-label? Maybe not, as it seems to be the only `<nav>` we have in our app, and it's likely self-explanatory.

frontend/src/components/HandbookOverlay/HandbookAttacks.tsx
```diff
@@ -22,10 +22,10 @@ function HandbookAttacks({ currentLevel }: { currentLevel: LEVEL_NAMES }) {
 
 	return (
 		<article className="handbook-page">
-			<header>
```
chriswilty, line 25: Interested to know why you changed this. MDN seems to indicate `<header>` is legitimate here:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header

frontend/src/components/HandbookOverlay/HandbookOverlay.css
```diff
@@ -1,55 +1,36 @@
...
-.handbook-overlay .close-button {
-	position: absolute;
-	top: 0.25rem;
-	right: 0.25rem;
-}
-
-.handbook-overlay .close-button:hover svg rect {
-	fill: black;
-	stroke: white;
+.handbook-overlay header {
```
chriswilty, line 8: Oh, now I can guess why you changed the other header into a div! This is quite a loose css selector, so all header elements within the overlay would've received these styles.

In this case, it makes more sense to use the classname as selector:
`.handbook-overlay .overlay-header`

frontend/src/components/Overlay/OverlayHeader.css
```diff
@@ -0,0 +1,41 @@
...
+}
+
+.overlay-header button:focus-visible {
+	background-color: var(--overlay-tab-colour);
+	outline-color: var(--main-text-colour);
+}
+
+.overlay-close-icon {
+	width: 1.2rem;
+	height: 1.2rem;
```
chriswilty, line 40: `1.25rem`

frontend/src/components/Overlay/OverlayHeader.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+		expect(screen.getByText('Test Header')).toBeInTheDocument();
+	});
+
+	test('renders close button', () => {
+		render(
+			<OverlayHeader heading="Test Header" closeOverlay={mockCloseOverlay} />
+		);
+
+		expect(screen.getByText('close')).toBeInTheDocument();
+	});
```
chriswilty, line 23: There's one more test you could add for the sake of completion: when you click the close button, the callback is fired.

For that you'll need user-event: see LevelSelectionBox for an example, though note you'll only need it for this new test, so you can use a raw `await userEvent.click()` within the test itself, without needing to do `userEvent.setup()`. Ask if you're unsure.

frontend/src/components/Overlay/OverlayHeader.tsx
chriswilty, line 16: In fact, rather than declaring this as a separate const, better to use a default value for the prop itself:

```
function OverlayHeader({
  closeOverlay,
  heading,
  iconColor = '#000',
}: OverlayHeaderProps) {
```

chriswilty (changes requested):
Excellent work, looks amazing ❤️ 

Just a handful of minor suggestions.

frontend/src/components/DocumentViewer/DocumentViewBoxNav.tsx
dhinrichs-scottlogic: The first rule of aria: Don't use aria if you don't have to. 
Based on the context (the h1 of the container, the nav element and the buttons), it's clear what the use is, so no aria required.

frontend/src/components/HandbookOverlay/HandbookAttacks.tsx
dhinrichs-scottlogic: During development I was trying to avoid confusing myself with the headers since the OverlayHeader is a header, but you're right, I'll change it back to a header. 
It also adds a nice visual presentation:
image

dhinrichs-scottlogic: yep, will change it back to header and use the class selector instead for the styling of the first header

frontend/src/components/DocumentViewer/DocumentViewBoxNav.tsx
dhinrichs-scottlogic: yup, just checked with NVDA on Firefox, it announces "navigation landmark" nicely 😎

dhinrichs-scottlogic:
@pmarsh-scottlogic Chris did some css magic, the issue with the scrollbar and margins should be fixed and future-proofed now. 
image

chriswilty (approved):
Lovely ❤️

Issue #777, opened 2024-01-16
Update level button styles to clarify their state


chriswilty:
Level buttons currently look like this:
![Image](https://github.com/ScottLogic/prompt-injection/assets/15246391/b4bbd369-204c-4ff5-a293-91f66fa65310)

It is not immediately obvious in this case if we are on Level 1 or Sandbox.

Following a discussion on Teams, we decided to tweak the styling a little, to make the distinction clearer:
![Image](https://github.com/ScottLogic/prompt-injection/assets/15246391/b55d6445-0273-438c-9602-a2052e46693e)

- Selected level has white background with text color the same teal as the header action button background-color
- Non-selected, non-disabled levels have same teal background as the header action buttons
- Disabled buttons are unchanged
- Focus styles are unchanged
- Hover styles should be removed from the selected level button, and use default cursor, to indicate the button doesn't do anything on click. It is not disabled, it just doesn't do anything as we are already on that level.

PR #831, opened 2024-02-09
777 changes to css to change button color on selected/hover

AAloshine-scottlogic:
## Description

changed level buttons/sandbox button to be more in line with the design of the rest of the app.

## Screenshots
image
image

## Checklist
https://github.com/ScottLogic/prompt-injection/issues/777

### DIFF
--- a/frontend/src/components/LevelSelectionBox/LevelSelectionBox.css
+++ b/frontend/src/components/LevelSelectionBox/LevelSelectionBox.css
@@ -21,6 +21,7 @@
 
 .level-button.selected {
 	background-color: var(--level-button-background-colour-selected);
+	cursor: default;
 	color: var(--level-button-colour);
 }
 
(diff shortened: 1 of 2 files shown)


chriswilty (approved):
Looks good to me!

frontend/src/components/LevelSelectionBox/LevelSelectionBox.css
```diff
@@ -22,6 +21,7 @@
 
 .level-button.selected {
 	background-color: var(--level-button-background-colour-selected);
+	color: var(--level-button-colour);
```
chriswilty, line 24: Actually, I wonder if we could put `cursor: default` here on the selected level, instead of using pointer, to make it clearer that it's not actionable when the level is already selected.

chriswilty (changes requested):
Oh one suggestion, sorry!

chriswilty (approved):
Lovely 💎

Issue #809, opened 2024-02-01
Setup Route 53 for Application Load Balancer


chriswilty:
Following on from #806, we will want to connect the Application Load Balancer with Route 53, so that UI and API are on the same domain. We could then disable CORS entirely.

Strictly speaking this is optional, as we could get away with a secure, cross-site cookie, but it will be nice to complete the job.

PR #834, opened 2024-02-13
385: Put APIGateway in front of private loadbalancer

### LINKED ISSUE #385: Fargate/ECS stack for back-end deployment

chriswilty:
## Description

Before we start using our spylogic.ai domain, I wanted to be able to stand up the application and test front-to-back, particularly given the difficulties around configuring secure cookies. For this, API Gateway seemed ideal... There were a few hiccups along the way, and some Express wrangling was needed (which can hopefully be removed once we are using spylogic.ai) but it all appears to work - remotely and locally.

Resolves #385 

## Screenshots

I may as well put a screenshot here of it deployed in AWS... Can do that once our new AWS account is up and running.

## Notes

- There are a few things still to cleanup, will do that shortly - hence PR is a draft for now.
- I've tested that it still works locally, but worth testing your own local too.
- I will close #385 after merging this PR, and handle reverting the API Gateway stuff and making our Application Load Balancer secure in #809

## Checklist

Have you done the following?

- [ ] Added tests


backend/src/app/sessionRoutes.ts
```diff
@@ -46,52 +50,51 @@ if (!sessionSigningSecret) {
...
 
-router.use((req, _res, next) => {
-	if (!req.session.initialised) {
-		req.session.chatModel = defaultChatModel;
-		req.session.levelState = getInitialLevelStates();
-		req.session.initialised = true;
-	}
-	next();
-});
+// TODO: Remove this debug logging!
```
chriswilty, line 88: Going to leave this in for now.

cloud/lib/api-stack.ts
```diff
@@ -104,6 +134,57 @@ export class ApiStack extends Stack {
...
+		});
+		vpcLinkSecurityGroup.connections.allowFromAnyIpv4(
+			Port.tcp(80),
+			'APIGW to VPCLink'
+		);
+		vpcLinkSecurityGroup.connections.allowTo(
+			fargateService.loadBalancer,
+			Port.tcp(80),
+			'VPCLink to ALB'
+		);
```
chriswilty, line 152: SecurityGroups are what allow/disallow access to resources inside a VPC (virtual private cloud).

VPCs come with a default security group allowing pretty much everything inbound and outbound, so for obvious reasons those rules are removed (CDK does this for us, which is nice). Annoyingly, the default security group itself cannot be removed, just the rules 🙄 

Here we add a security group to our VPCLink, to allow inbound TCP traffic on port 80 (from API Gateway) and outbound traffic to our Application Load Balancer only.

backend/src/app/nonSessionRoutes.ts
chriswilty, line 1: I moved the routing and express app into its own folder, as it was all getting a bit noisy in src.

backend/src/app/sessionRoutes.ts
```diff
@@ -8,34 +8,38 @@ import {
...
+import { LevelState, getInitialLevelStates } from '@src/models/level';
 
 declare module 'express-session' {
 	interface Session {
 		initialised: boolean;
 		chatModel: ChatModel;
 		levelState: LevelState[];
 	}
+	interface CookieOptions {
+		// TODO express-session types are not yet up-to-date!
```
pmarsh-scottlogic, line 40: I'm guessing you're saving this for a future PR?

backend/src/app/sessionRoutes.ts
```diff
@@ -46,52 +50,53 @@ if (!sessionSigningSecret) {
 	process.exit(1);
 }
 
-const router = express.Router();
-
+const cookieName = process.env.COOKIE_SID;
```
pmarsh-scottlogic, line 53: We need to add a new thing to .env? Shall we add something to .env.example too?

pmarsh-scottlogic: nvm, I've seen that it gets added to the fargate instance, and it runs fine locally

pmarsh-scottlogic:
by the way, looks like you can merge this without approval. I didn't know it that's what you wanted for the branch protections

pmarsh-scottlogic (approved):
all looks sensible

backend/src/app/sessionRoutes.ts
chriswilty: Ah yes, it falls back to the default cookie name, but it would probably be clearer to use our own specific cookie name, so I'll add it to the example env 👍

Issue #699, opened 2023-12-20
Adjust header and footer for text resizing [a11y]


dhinrichs-scottlogic:
# Feature Request

## Description
Adjust header and footer so that the text can be resized to 200% without losing functionality. 

- Assume that you're using the app full screen on 13" laptop at 1920x1080 resolution
- This ticket requires some research to familiarise yourself with the sufficient and advisory techniques for meeting this accessibility criterion. 

## Additional context
This is to fulfil WCAG 1.4.4 Resize Text.  https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html
This means it should be possible to increase the font-size to 200% and no content should be cut off or hidden. 

### implementation details
some ideas for how this can be done:
- reduce overall header size (a height of 5rem might be enough?) and maybe give it a max-height to give more space to the main page, or, use media queries to reduce header height at a specific width breakpoint.
- give the Spy Logic logo a max-width or max-height
- reduce some of the gaps and padding
- allow the buttons to wrap their text?
- remove icons from buttons at a small breakpoint
- reduce footer height; this could be as little as 2rem.

## Screenshots

How it looks currently at 200%
![image](https://github.com/ScottLogic/prompt-injection/assets/15246391/1a7b7154-dad1-4f2b-a462-5fdba5fc5465)

Roughly expected outcome at 100% (no zoom)
![Image](https://github.com/ScottLogic/prompt-injection/assets/125262707/37890ff7-a210-4830-8192-b6fbb516dbf5)

Roughly expected outcome at 200% 
![Image](https://github.com/ScottLogic/prompt-injection/assets/125262707/86621e0e-316a-4c30-b8bf-86247dede9c1)

## Acceptance criteria
**GIVEN** the user uses the app in the full screen browser on a 13" laptop
**WHEN** the user zooms to 200%
**THEN** all of the buttons and links of the header and footer are still visible.

Kenny4297:
Hi @dhinrichs-scottlogic, I believe I can help with this issue. Would you mind assigning it to me?

dhinrichs-scottlogic:
> Hi @dhinrichs-scottlogic, I believe I can help with this issue. Would you mind assigning it to me?

Yes, go ahead, I've assigned this issue to you. Thank you so much for contributing!

Kenny4297:
[...]

PR #835, opened 2024-02-14
adding formatting for 200% magnification on standard screen

AAloshine-scottlogic:
## Description
added media break querys to decrease padding and margins so that the app fully displays all functions at 200% magnification. 

## Screenshots

image

ticket link:
https://github.com/ScottLogic/prompt-injection/issues/699


dhinrichs-scottlogic (changes requested):
Here's what I found so far:
- On Chrome, when you magnify to 150% the Getting Started and Reset Progress buttons touch
- Could you move the Level 1 2 3 section further to the left so that the Reset Progress doesn't wrap?

chriswilty (changes requested):
This isn't working as you intended, since switching to rems, so I assume you didn't run it locally to check

The logic is basically inverted now at normal (100%) zoom, as in, the "render everything smaller and tighter" breakpoint kicks in _from_ 600px width and upwards:

At 599px:
image

At 600px:
image

At 1480px (laptop screen):
image

And at 200% zoom on full laptop screen it still doesn't all fit:
image

If I rewind to your original code using 150DPI, it doesn't work well on my screen where I have my taskbar vertically:

image

And of course with DPI we don't get any allowance for (available) screen width nor basefont increases.

I think you'd be wise to start from scratch, and work out a breakpoint (in rem) where things become too tight, on a standard ScottLogic laptop (1920x1200). If need be, you can gracefully respond using a couple of breakpoints, though that might be beyond the scope of this ticket if you can get everything working nicely with a single change. Remember to test it with
- zooming in/out, up to 200%
- changing base font at 100% zoom only, up to 32px
- Resizing the window

AAloshine-scottlogic:
Ah crumbs, youre right! i've been using the desktop to test the magnification but havent used my laptop since i usually have my code on the laptop screen and the website open on the desktop one. I'll go ahead and try and figure out the correct rem for the laptop screen then

chriswilty:
This is looking really good now 💎 

Just the minor suggestions I chucked your way on teams, then very happy to approve ✅

chriswilty (approved):
Lovely!

Issue #824, opened 2024-02-06
Streamline frontend change level network calls 🛠🛠🛠


pmarsh-scottlogic:
Here's what happens when I change from sandbox to level 1:

- `email/get?level=0` returns all the sent emails in the session for given level.
- `openai/history?level=0` returns the chat history in the session for given level.
- `defence/status?level=0` returns the defences as configured for given level.

These three are also called when you refresh.

🔧 These should be combined.

# AC:

GIVEN frontend in preview mode (so that react doesn't do its double render thing) `npm run preview`
WHEN the user changes from one level to another (check specifically switching in and out of sandbox)
THEN there should be only one api call made of type fetch

Here's how you quickly find out what api calls are being made that are fetch type. Note that the CombinedFonts call is of type xhr, therefore we ignore it.

![image](https://github.com/ScottLogic/prompt-injection/assets/118171430/48b15e2b-5cbe-4e1c-9495-8848d400c03c)

PR #836, opened 2024-02-14
300 streamline frontend refresh network calls

pmarsh-scottlogic:
## Description

When you first load the website in the browser, or refresh we only send one network request which should grab all the necessary information to get going.

## Notes

- Creates `GET /start` endpoint which is the single network call the frontend needs to set itself up. It returns the available models, the system roles for all levels and the emails, history and defences for the current level.
- The above network call acts as the frontend's health check so I've removed the healthcheck stuff from the frontend
- removes `GET /openai/validModels` endpoint and all related code (no longer needed)
- removes `GET /systemRoles` endpoint and all related code (no longer needed)
- moves `loadBackendData` and `openHandbook` to _mainComponent.tsx_ so that all the `/start` fetched state stuff can be managed by one component.
- Adds `UseIsFirstRender` hook, for using with the useEffect hook. It indicates whether the useEffect is running on the first mount or on the change of a dependency.

## Checklist

Have you done the following?

- [ ] Added tests


chriswilty:
@pmarsh-scottlogic Reviewing shortly, lunch first...

frontend/src/hooks/useIsFirstRender.ts
```diff
@@ -0,0 +1,11 @@
+import { useRef, useEffect } from 'react';
+
+function UseIsFirstRender() {
+	const isMountRef = useRef(true);
+	useEffect(() => {
+		isMountRef.current = false;
+	}, []);
+	return isMountRef.current;
+}
```
chriswilty, line 9: Feels a bit overkill to extract this into its own hook. All you're really doing is storing a ref to indicate when the component has mounted.

frontend/src/models/combined.ts
chriswilty, line 1: "combined" ... Is this all the data we expect in the session? Maybe currentState is a touch clearer?

frontend/src/service/startService.ts
```diff
@@ -0,0 +1,23 @@
...
+		method: 'GET',
+	});
+	const startResponse = (await response.json()) as StartReponse;
+
+	return {
+		emails: startResponse.emails,
+		history: getChatMessagesFromDTOResponse(startResponse.history),
+		defences: startResponse.defences,
+		availableModels: startResponse.availableModels,
+		systemRoles: startResponse.systemRoles,
```
chriswilty, line 19: Can probably just destructure all these.

```
const { availableModels, defences, emails, history, systemRoles } = (await response.json()) as StartResponse;
```

frontend/src/components/MainComponent/MainComponent.tsx
```diff
@@ -116,21 +150,27 @@ function MainComponent({
 
 	// for going switching level without clearing progress
 	async function setNewLevel(newLevel: LEVEL_NAMES) {
-		// get emails for new level from the backend
-		setEmails(await emailService.getSentEmails(newLevel));
+		const emails = await emailService.getSentEmails(newLevel);
+		const chatHistory = await chatService.getChatHistory(newLevel);
+		const defences = await defenceService.getDefences(newLevel);
```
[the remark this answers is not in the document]
pmarsh-scottlogic: yep. #824

backend/src/controller/startController.ts
```diff
@@ -0,0 +1,30 @@
...
+} from '@src/promptTemplates';
+
+function handleStart(req: GetStartRequest, res: Response) {
+	const level = req.query.level;
+
+	const systemRoles = [
+		{ level: LEVEL_NAMES.LEVEL_1, systemRole: systemRoleLevel1 },
+		{ level: LEVEL_NAMES.LEVEL_2, systemRole: systemRoleLevel2 },
+		{ level: LEVEL_NAMES.LEVEL_3, systemRole: systemRoleLevel3 },
+	];
```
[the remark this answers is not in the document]
chriswilty: I don't think that's complex, as we are sending a payload back to the UI when the level is beaten anyway. It seems silly to hide the system role in the handbook, but it's right there in the network tab if you go looking.

frontend/src/components/MainComponent/MainComponent.tsx
```diff
@@ -116,21 +150,27 @@ function MainComponent({
...
+		level !== LEVEL_NAMES.SANDBOX
+			? setMessagesWithWelcome(chatHistory)
+			: setMessages(chatHistory);
 
 		const defences =
-			newLevel === LEVEL_NAMES.LEVEL_3 ? DEFENCES_SHOWN_LEVEL3 : ALL_DEFENCES;
-		// fetch defences from backend
-		const remoteDefences = await defenceService.getDefences(newLevel);
+			level === LEVEL_NAMES.LEVEL_3 ? DEFENCES_SHOWN_LEVEL3 : ALL_DEFENCES;
 		defences.map((localDefence) => {
```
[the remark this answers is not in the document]
pmarsh-scottlogic: Yup, removed the mutations, and also moved all the logic to the service, so that the service returns everything we need to just set it directly

frontend/src/hooks/useIsFirstRender.ts
pmarsh-scottlogic: I see no reason to not extract self-contained logic, even if it is trivial. Especially if un-extracting it will take extra effort. Do you think it is less readable this way?

frontend/src/models/combined.ts
pmarsh-scottlogic: It is not. We are missing out the chatModel property. And systemRole, availableModels are not in the backend session.

I called it combined because it is just miscellaneous combinations of other things. I don't think it's a good name, but I can't think of anything better. When I implement #842 I'll put those types in here as well.

I suppose I could call it DTOs?

frontend/src/hooks/useIsFirstRender.ts
chriswilty: There are plenty of reasons not to extract self-contained logic, particularly into its own file. This one is used in just one component, which is my primary argument. It is no _more_ readable sitting in its own file, than existing as a useRef in the same component that needs it.

Fragmenting your code is something to do when it aids in understanding and readability, because your code is becoming complex. But even in that case, there is rarely much benefit in extracting small pieces of code; instead you will benefit most from separating concerns.

I would like you to learn that refactoring can make things worse; keep related code co-located until such time as you find yourself repeating / copying logic. I am a strong advocate for keeping things simple; the Three Strikes rule is a good one to follow in cases like this:
https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

frontend/src/models/combined.ts
chriswilty: No biggie, let's see if something else emerges in time!

frontend/src/components/ChatBox/ChatBox.tsx
```diff
@@ -111,7 +111,7 @@ function ChatBox({
 		}
 		response.defenceReport.alertedDefences.forEach((triggeredDefence) => {
 			// get user-friendly defence name
-			const defenceName = ALL_DEFENCES.find((defence) => {
+			const defenceName = DEFAULT_DEFENCES.find((defence) => {
```
chriswilty, line 114: Ah, interesting... Is this because System Role and the Q&A LLM are not defences that are "triggered" like the others? If so, this provides useful clarity 👍

frontend/src/components/ControlPanel/ControlPanel.tsx
```diff
@@ -29,12 +29,21 @@ function ControlPanel({
...
+			: [];
+
+	const nonModelDefences = configurableDefences.filter(
+		(defence) => !MODEL_DEFENCES.includes(defence.id)
 	);
 
-	const modelDefences = defences.filter((defence) =>
-		MODEL_DEFENCES.some((id) => id === defence.id)
+	const modelDefences = configurableDefences.filter((defence) =>
+		MODEL_DEFENCES.includes(defence.id)
```
chriswilty, line 46: Filtering / iterating twice seems unnecessary; I came up with a reduce to split it over one iteration but that felt really clunky. This is one area where javascript feels a bit weak... Small beans, leave it!

chriswilty (commented):
This all seems great 💎 

Still the reservation about extracting the hook, otherwise very happy to approve.

frontend/src/hooks/useIsFirstRender.ts
pmarsh-scottlogic: Interesting... This feels like something I'd have to learn the hard way to properly internalise it, but I'll take it on face value for now!

chriswilty: I _have_ learned the hard way ;)

chriswilty: _(Chris keeps quietly spreading the word about ReScript)_

chriswilty (approved):
Very happy with this, took some working out 🏆

frontend/src/components/ChatBox/ChatBox.tsx
pmarsh-scottlogic: True they're not "triggered", but then also the message transformation defences aren't triggered either.
It's mostly because those two aren't configurable in level 3

PR #838, opened 2024-02-16
578 implement multi page modals

dhinrichs-scottlogic:
## Description
Modals now new with multiple pages and fancy character images!

Closes 578

## Screenshots
image

## Notes
- modals zoom up to 400%

## Concerns
- unsure if I've implemented the image src prop passing correctly
- I tried to reduce the number of media query breakpoints, but then the zooming didn't work anymore, so reverted to a version that worked

## Checklist

Have you done the following?

- [ ] Added tests


frontend/src/components/Overlay/OverlayHeader.css
```diff
@@ -3,20 +3,22 @@
 	justify-content: space-between;
 	align-items: center;
 	height: 5rem;
+	margin: 0;
+	padding: 0 1em;
```
chriswilty, line 7: `rem` not `em`?

frontend/src/components/Overlay/OverlayHeader.css
```diff
@@ -3,20 +3,22 @@
...
-	padding: 0 2rem;
+	margin: 0;
 }
 
 .overlay-header button {
 	display: flex;
 	gap: 0.5rem;
 	align-items: center;
-	margin-right: 1rem;
+	margin: 0;
```
chriswilty, line 21: Buttons are `margin: 0` by default, unless we are inheriting margin from somewhere else.

frontend/src/components/Overlay/OverlayHeader.css
```diff
@@ -39,3 +41,36 @@
...
+}
+
+@media (width < 62.5rem) {
+	.overlay-header {
+		height: auto;
+		padding: 0 0.125rem;
+	}
+
+	.overlay-header h1 {
+		border-top-left-radius: 0.625rem;
```
chriswilty, line 64: Hmm, interesting. Is the header overflowing the header boundaries without this? Feels a bit strange to need a single corner radius which needs to match some other radius, instead of using e.g. overflow: hidden.

frontend/src/components/Overlay/OverlayNav.css
```diff
@@ -0,0 +1,92 @@
+.overlay-nav {
+	display: grid;
+	grid-template-columns: 1fr 1fr 1fr;
+	gap: 1rem;
+	height: 5rem;
+	padding: 0;
+	border-bottom-right-radius: 0.625rem;
+	border-bottom-left-radius: 0.625rem;
```
chriswilty, line 8: Again, does this footer region overflow the overlay itself, or why did you need explicit radius here?

frontend/src/components/Overlay/OverlayNav.css
```diff
@@ -0,0 +1,92 @@
...
+	background-color: var(--overlay-header-background-colour);
+}
+
+.overlay-nav div {
```
chriswilty, line 12: This is quite a loose selector, is it what you intended - for all divs inside the nav to be centered flex containers?

Looking at the markup, it feels like the two divs wrapping the buttons are unnecessary - can we not just have buttons, assigned to and aligned within their respective grid slots?

frontend/src/components/Overlay/OverlayNav.tsx
```diff
@@ -0,0 +1,53 @@
...
+					</button>
+				)}
+			</div>
+			<p>{pagination}</p>
+			<div className="next-container">
+				{!nextDisabled && (
+					<button
+						onClick={goToNextPage}
+						aria-disabled={nextDisabled}
+						className="nav-buttons next"
```
chriswilty, line 40: This button has className "next" which allows us to apply styles to it, such as `justify-self: end` and `grid-column: 3`. Can do the same for the "previous" button, then there should be no need for those wrapper divs.

Also, `aria-disabled` will always be false here.

frontend/src/components/ThemedButtons/OverlayButton.css
```diff
@@ -2,13 +2,14 @@
 	display: flex;
 	justify-content: center;
 	align-items: center;
+	align-self: center;
+	width: fit-content;
```
chriswilty, line 6: Are you sure you want these? This button is always going to override align-items on the parent... Is it deliberate, so that you don't need to use align-items on every parent container that uses this button?

chriswilty (changes requested):
Some minor suggestions and css questions, but it's all looking really nice 👌

frontend/src/components/Overlay/LevelsComplete.tsx
```diff
@@ -10,35 +9,40 @@ function LevelsComplete({
...
+		</>,
+		<>
+			<h2>You&apos;ve completed the story mode</h2>
+			<p>
+				You can stay here and continue to play with the levels, or you can move
+				onto Sandbox mode where you can configure your own defence set up and
+				try to break it.
+			</p>
+			<p>
+				You can always switch modes by clicking on the button in the left panel.
```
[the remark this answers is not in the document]
dhinrichs-scottlogic: It was about switching back and forth between sandbox and levels if I recall correctly. I didn't make any changes to the text, it's leftover from earlier versions of the game. 

Maybe that paragraph ("You can always switch modes by clicking on the button in the left panel.") can be entirely removed, since by the time you get the win modal, users will be familiar with the UI, the level buttons and the Getting Started button?

frontend/src/components/Overlay/OverlayHeader.css
dhinrichs-scottlogic: ah, yep, that's not needed, that was from development, if you give the header a background colour you see if overflowing, but that's obviously not an issue if the header does not have a background color. Will remove it.   
image

frontend/src/components/Overlay/OverlayNav.css
dhinrichs-scottlogic: The OverlayNav is not part of the Overlay component, so I do think this is needed, but happy to change it if there's a better way? 
Without: 
image

frontend/src/components/Overlay/OverlayNav.css
dhinrichs-scottlogic: Since there are only two divs and I wanted to target those two, yes that was intended, but I can see how this could be accidentally targeting more divs if we add more divs, happy to add another class to make it more specific. 

The divs properly center the buttons, without it looks like this: 
image
This is working quite well, including with zoom, don't want to fix it if it isn't broken.

frontend/src/components/Overlay/MultipageOverlay.tsx
```diff
@@ -39,16 +39,14 @@ function MultipageOverlay({
...
 		speakerImage = Manager;
-	} else if (imgSource[currentPage] === 'ScottBrew Lawyer') {
+	} else if (pages[currentPage].imageName === 'ScottBrew Lawyer') {
 		speakerImage = Lawyer;
-	} else if (imgSource[currentPage] === 'SpyLogic') {
-		speakerImage = BotAvatarDefault;
 	} else {
-		speakerImage = '';
+		speakerImage = BotAvatarDefault;
 	}
```
chriswilty, line 50: The only thing I'd consider changing here is forcing the components that are using MultipageOverlay to import their own images and pass them in, instead of passing in a name and using this if...else block to pick the correct image.

The reasoning being, now MultipageOverlay needs knowledge of all the images that could possibly be requested by users of this component, so if you want a new image, you'd need to make code changes here and in the parent component.

frontend/src/components/Overlay/OverlayNav.css
chriswilty: Right that makes sense!

chriswilty (commented):
One suggestion, to decouple the MultipageOverlay from our image files, else we're good to go!

Hmm, it's actually above this review comment 🤔 
https://github.com/ScottLogic/prompt-injection/pull/838#discussion_r1504125732

frontend/src/components/Overlay/MultipageOverlay.tsx
dhinrichs-scottlogic: ah yes that makes sense, didn't even know that would work but it does :)

chriswilty: It works because when you import an image file in Vite, you actually get a URL to that image, which is a string. The image itself is a static asset:
https://vitejs.dev/guide/assets#static-asset-handling

For this reason, it might be clearer to call the prop "imageUrl" rather than "imageName".

chriswilty (approved):
🏆

PR #839, opened 2024-02-22
824 streamline frontend change level network calls

pmarsh-scottlogic:
## Description

combines the change level network calls into one

## Notes

- makes new endpoint /level which returns the emails, chat history and defences for a given level
- deletes the now unused endpoints that get the email, chat history and defences individually
- renames the existing `levelService` to `resetService` (because all it was doing was resetting a level)
- adds new `levelService` for accessing our new endpoint

### Further PRs

- I notice that we pass the entire defences object to the front end when switching to and loading up levels 1 and 2. However we don't use any of this information unless we're showing it in the ControlPanel. Which is both confusing to read and also a waste of bandwidth. To fix this I think would warrant a separate PR.
- There is also the problem of having lots of getModel calls when we switch to sandbox. That can also be a separate PR

## Checklist

Have you done the following?

- [ ] Added tests


backend/src/controller/startController.ts
```diff
@@ -9,7 +9,7 @@ import {
 	systemRoleLevel3,
 } from '@src/promptTemplates';
 
-function handleStart(req: GetStartRequest, res: Response) {
+function handleStart(req: StartGetRequest, res: Response) {
```
pmarsh-scottlogic, line 12: renamed this to be more consistent with the other model names

frontend/src/service/levelService.ts
pmarsh-scottlogic, line 1: this diff is all screwed up.
What's happened is that I've moved all the resetAllLevelProgress logic to a new file resetService, then put the loadLevel logic into this file

backend/test/unit/controller/chatController.test.ts
```diff
@@ -741,55 +739,6 @@ describe('handleChatToGPT unit tests', () => {
 	});
 });
 
-describe('handleGetChatHistory', () => {
```
chriswilty, line 744: As you've combined the logic into levelController, and deleted these tests, shouldn't you be replacing them with a test suite for the new controller, rather than just deleting them?

frontend/src/App.tsx
```diff
@@ -7,7 +7,7 @@ import MissionInformation from './components/Overlay/MissionInformation';
 import OverlayWelcome from './components/Overlay/OverlayWelcome';
 import ResetProgressOverlay from './components/Overlay/ResetProgress';
 import { LEVEL_NAMES } from './models/level';
-import { levelService } from './service';
+import { resetService } from './service';
```
chriswilty, line 10: Yep that's fine, and I assume when you get around to tackling the reset level call, it will go in resetService as well?

chriswilty (changes requested):
I'd ideally like to see replacement tests rather than just deleting the old ones, but if you think it's really not worth the effort (as it's only really extracting stuff from local storage and sending it in a response) then I can live with that.

I am far more concerned with our lack of UI component tests in general, which is not something you can or should tackle here!

frontend/src/App.tsx
pmarsh-scottlogic: that's that plan!

backend/src/controller/handleError.ts
```diff
@@ -22,7 +22,8 @@ function handleChatError(
 		reply: errorMsg,
 		isError: true,
 	};
-	res.status(statusCode).send(updatedChatResponse);
+	res.status(statusCode);
+	res.send(updatedChatResponse);
```
chriswilty, line 26: What's the difference here?

backend/test/unit/controller/levelController.test.ts
```diff
@@ -0,0 +1,92 @@
...
+		chatModel: ChatModel;
+		levelState: LevelState[];
+	}
+	interface LevelState {
+		level: LEVEL_NAMES;
+		chatHistory: ChatMessage[];
+		defences: Defence[];
+		sentEmails: EmailInfo[];
+	}
+}
```
chriswilty, line 24: Hmm, interesting. This is now replicated in implementation and test code, so maybe we need to be extracting the typedef into a shared location that both tsconfig can refer to. Feel free to tackle that later though.

backend/test/unit/controller/levelController.test.ts
```diff
@@ -0,0 +1,92 @@
...
+		send: mockSend,
+		status: jest.fn().mockReturnValue({ send: mockSend }),
+	} as unknown as Response;
+}
+
+afterEach(() => {
+	mockSend.mockClear();
+});
+
+test('WHEN client asks to load level 1 THEN the backend sends the level information for the given level', () => {
```
chriswilty, line 45: So the level number itself is unimportant? Or is it that the levelState array only has one entry in it, hence we return the first entry for level 1? If the latter is true, then it'd be good to have tests ensuring all our levels are covered so we know we are returning the correct level state, or at least that level1 and sandbox are returned, and also that a level outside of bounds returns a 400.

backend/test/unit/controller/startController.test.ts
```diff
@@ -0,0 +1,99 @@
...
+		send: mockSend,
+		status: jest.fn().mockReturnValue({ send: mockSend }),
+	} as unknown as Response;
+}
+
+afterEach(() => {
+	mockSend.mockClear();
+});
+
+test('GIVEN level 1 provided WHEN user starts the frontend THEN the backend sends the initial information', () => {
```
chriswilty, line 45: I'm going to suggest the same about this test suite as well - cover first and last level ok, then cover level outside of bounds returns 400.

frontend/src/service/levelService.ts
```diff
@@ -1,15 +1,23 @@
+import { LoadLevelResponse } from '@src/models/combined';
+
 import { sendRequest } from './backendService';
+import { getChatMessagesFromDTOResponse } from './chatService';
+import { getDefencesFromDTOs } from './defenceService';
 
-const PATH = 'reset';
+const PATH = 'level/';
```
chriswilty, line 7: Intentional trailing slash? Seems wrong, even though express lets you get away with it on the backend.

chriswilty (changes requested):
Looking good 👍 

Handful of minor suggestions.

backend/src/controller/handleError.ts
pmarsh-scottlogic: oh yeah, I did that while experimenting with mocking, and forgot to change it back

backend/test/unit/controller/chatController.test.ts
pmarsh-scottlogic: done

backend/src/controller/handleError.ts
pmarsh-scottlogic: reverted

backend/src/controller/startController.ts
```diff
@@ -11,6 +12,14 @@ import {
 
 import { sendErrorResponse } from './handleError';
 
+declare module 'express-session' {
+	interface Session {
+		initialised: boolean;
+		chatModel: ChatModel;
+		levelState: LevelState[];
+	}
+}
```
chriswilty, line 21: In how many places do we have this module declaration now? We need to fix it...

chriswilty: Well, looks like we've fixed it by using a typings folder 🥳

backend/test/unit/controller/levelController.test.ts
```diff
@@ -43,16 +92,13 @@ test('WHEN client asks to load level 1 THEN the backend sends the level informat
...
-		emails: [],
-		chatHistory: [],
-		defences: [],
-	});
+	expect(res.status).toHaveBeenCalledWith(400);
+	expect(mockSend).toHaveBeenCalledWith('Level not provided');
 });
 
-test('WHEN client does not provide a level THEN the backend sends the level information for the given level', () => {
+test('WHEN client provides an invalid level THEN the backend sends the level information for the given level', () => {
```
chriswilty, line 99: `THEN the backend responds with BadRequest` ?

chriswilty (approved):
Beautiful! One test name looking like it needs correcting, else good to go.

PR #841, opened 2024-02-23
Make short mission info non-clickable and add button #535

### LINKED ISSUE #535: Make short mission info non-clickable and add button
Currently the Short Mission Info banner is a clickable button. This is not super clear to visual users and gives a super long text for screen readers. 

The plan:
- Turn the banner into a non-clickable element.
- Add a button on the right that lets users bring up the full mission info again.
- Turn the "Level 1" heading into an h2
- Ensure the icon on the button is aria-hidden

image

Acceptance Criteria:
**GIVEN**  a visual user is going through the Level mode
**WHEN** they look at the banner
**THEN** they see a short mission info on the banner and a button which brings up the full mission info

**GIVEN**  a screen reader user is going through the Level mode
**WHEN** they interact with the banner
**THEN** the "Level 1" on the banner is an h2 heading, the short mission info is a text paragraph, and there is a clickable button which announces the button text but not the icon.

AAloshine-scottlogic:
## Description

changed level banner format to include actual button with mission info as per ticket #535 

## Screenshots

image

## Notes

- Bullet point list with
- any notable code changes

|## link:
https://github.com/orgs/ScottLogic/projects/37/views/2?pane=issue&itemId=44015553

Resolves #535

### DIFF
--- a/frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
+++ b/frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
@@ -3,7 +3,6 @@
 	gap: 0.5rem;
 	align-items: center;
 	padding: 0.5rem 1.5rem;
-	border: 0.5rem solid transparent;
 	background-color: var(--level-banner-background-colour);
 	color: var(--level-banner-text-colour);
 	text-align: left;
--- a/frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
+++ b/frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
@@ -26,6 +26,20 @@
 	font-size: 1rem;
 }
 
-.level-mission-info-banner:hover {
-	background-color: var(--level-banner-background-colour-hover);
+.level-mission-info-banner .themed-button {
+	height: auto;
+	white-space: nowrap;
+}
+
+.info-icon {
+	margin-right: 0.25rem;
+	padding: 0 0.4rem;
+	border: 0.2rem;
+	border-style: solid;
+	border-color: var(--action-button-text-colour);
+	border-radius: 50%;
+	background-color: var(--action-button-text-colour);
+	color: var(--action-button-background-colour);
+	font: 1rem serif;
+	font-weight: bold;
 }
--- a/frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
+++ b/frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
@@ -16,6 +16,7 @@
 
 .level-title-area {
 	min-width: 10%;
+	margin: 1rem 0;
 	font-weight: 800;
 	font-size: 1.875rem;
 }
@@ -26,14 +27,18 @@
 }
 
 .level-mission-info-banner .themed-button {
+	gap: 0.25rem;
 	height: auto;
+	padding: 0.75rem 1rem;
 	white-space: nowrap;
 }
 
 .info-icon {
-	margin-right: 0.25rem;
-	padding: 0 0.4rem;
-	border: 0.2rem;
+	display: flex;
+	align-items: center;
+	box-sizing: border-box;
+	height: auto;
+	padding: 0 0.625rem;
 	border-style: solid;
 	border-color: var(--action-button-text-colour);
 	border-radius: 50%;
@@ -41,4 +46,5 @@
 	color: var(--action-button-background-colour);
 	font: 1rem serif;
 	font-weight: bold;
+	aspect-ratio: 1;
 }
--- a/frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
+++ b/frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
@@ -15,7 +15,6 @@
 }
 
 .level-title-area {
-	min-width: 10%;
 	margin: 1rem 0;
 	font-weight: 800;
 	font-size: 1.875rem;
--- a/frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
+++ b/frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
@@ -1,6 +1,6 @@
 .level-mission-info-banner {
 	display: flex;
-	gap: 0.5rem;
+	gap: 1.5rem;
 	align-items: center;
 	padding: 0.5rem 1.5rem;
 	background-color: var(--level-banner-background-colour);
@@ -19,6 +19,7 @@
 	margin: 1rem 0;
 	font-weight: 800;
 	font-size: 1.875rem;
+	white-space: nowrap;
 }
 
 .level-mission-info-banner > p {
@@ -27,7 +28,7 @@
 }
 
 .level-mission-info-banner .themed-button {
-	gap: 0.25rem;
+	gap: 0.5rem;
 	height: auto;
 	padding: 0.75rem 1rem;
 	white-space: nowrap;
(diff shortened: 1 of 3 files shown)


dhinrichs-scottlogic:
It looks like the mission info button adds padding to the banner, probably from the default padding in the themed-button class.
Could that be adjusted so that the banner continues to take up the same amount of space as before?
image

frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.tsx
```diff
@@ -11,14 +12,21 @@ function LevelMissionInfoBanner({
...
+			<h2 className="level-title-area">{`Level ${currentLevel + 1}`}</h2>
 			<p
+				data-testid="banner-info"
 				dangerouslySetInnerHTML={{
 					__html: LEVELS[currentLevel].missionInfoShort ?? '',
 				}}
 			></p>
-		</button>
+			<ThemedButton onClick={openOverlay}>
+				<p className="info-icon" aria-hidden="true">
```
pmarsh-scottlogic, line 24: a tiny semantic improvement would be to switch this `<p>` tag to a `<span>` given it isn't containing a chunk of text.

pmarsh-scottlogic (changes requested):
Great job. still looks a bit chonky - I suggest adding `margin: 0` to the `level-title-area` class

chriswilty (changes requested):
Almost there...

frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.tsx
AAloshine-scottlogic: okay sounds good

frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
```diff
@@ -26,6 +25,20 @@
...
 
-.level-mission-info-banner:hover {
-	background-color: var(--level-banner-background-colour-hover);
+.level-mission-info-banner .themed-button {
+	height: auto;
+	white-space: nowrap;
+}
+
+.info-icon {
+	margin-right: 0.25rem;
```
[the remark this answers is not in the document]
AAloshine-scottlogic: ah i was wondering what was the semantic difference between gap and margins, will fix that now

chriswilty:
@dhinrichs-scottlogic, @pmarsh-scottlogic The chonkiness is due to the h2 of "Level N" having margins, nothing to do with the button. Can of course whack the margin (on "level-title-area") down a little:

image

`margin: 1rem 0`

frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
```diff
@@ -26,6 +25,20 @@
...
+	padding: 0 0.4rem;
+	border: 0.2rem;
```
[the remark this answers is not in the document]
chriswilty: Ideally we would be setting consistent box-sizing on all elements via the rule `body * { box-sizing: border-box }`, however, that is likely to cause some disruption as the default "content-box" is being applied in most places. I should probably add an issue to fix that...

frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
```diff
@@ -26,6 +25,20 @@
 	font-size: 1rem;
 }
 
-.level-mission-info-banner:hover {
-	background-color: var(--level-banner-background-colour-hover);
+.level-mission-info-banner .themed-button {
+	height: auto;
+	white-space: nowrap;
```
chriswilty, line 30: You might need some explicit padding here; I used `0.75rem 1rem` in my trials, which looked ok.

frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.tsx
```diff
@@ -11,14 +12,21 @@ function LevelMissionInfoBanner({
 	openOverlay: () => void;
 }) {
 	return (
-		<button className="level-mission-info-banner" onClick={openOverlay}>
-			<span className="level-title-area">{`Level ${currentLevel + 1}`}</span>
+		<span className="level-mission-info-banner">
```
chriswilty, line 15: @dhinrichs-scottlogic Would we be better off using some sort of landmark for this banner, or is it self-explanatory?

I'm wondering if "section" or "complementary" are appropriate. I do realise `<section>` is a bit controversial...
https://developer.mozilla.org/en-US/blog/aria-accessibility-html-landmark-roles/

chriswilty (changes requested):
Minor tweaks suggested, and one comment for @dhinrichs-scottlogic that we don't need to address in this issue.

frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.tsx
dhinrichs-scottlogic: any semantic element is better than a `<span>`. 
If I follow [this flowchart](https://html5doctor.com/downloads/h5d-sectioning-flowchart.pdf),  `<article>` or `<section>` seems reasonable

dhinrichs-scottlogic: landmarks should be [used sparingly](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Region_role#accessibility_concerns). If we give this a section, with a header, a text paragraph and a button, I think there's enough context for it without giving it a landmark role, and you already can easily jump to its header. The first rule of aria is: don't use aria if you can avoid it, that's probably true for role declarations, too. 
mdn's comedians are putting it aptly here: 
image

dhinrichs-scottlogic: I vote for `<article>`

frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
```diff
@@ -19,6 +19,7 @@
 	margin: 1rem 0;
```
chriswilty, line 19: Need to remove `min-width: 10%` above.

frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.tsx
chriswilty: Well, that's a possibility. Another is that they are list items within a list inside the chat panel, which itself has role=log:
https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA23
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/log_role

The first link has exactly our use case, plus a working demo. I note once again that Narrator is not announcing "polite" additions to the conversation log, but NVDA is.

_Do we want to add this improvement as a separate issue?_

dhinrichs-scottlogic: I don't think that's our use case, the short mission info doesn't keep a log, or updates, it just contains different text between the levels, which is expected since we are changing context. The same is true for the handbook, control panel, level buttons... 
I'd say keep it as simple as possible, use either `<section>` or `<article>`, it's at least a little bit more semantic than `<span>`, but don't add any additional roles or aria for now.

frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
AAloshine-scottlogic: ah sorry misssed that bit

PR #842, opened 2024-02-23
convert the level enum

pmarsh-scottlogic:
## Description

The level being an enum was disrupting with my network calls ticket #824 

## Notes

- Tried switching the values so that sandbox had value 0, level 1 had value 1, level 2 had value 2, etc. This caused lots of problems because many things depend on what order things are declared in arrays, which is not great. But not worth the effort to fix right now

## Checklist

Have you done the following?

- [ ] Linked the relevant Issue | no real ticket for this
- [ ] Added tests | no functional change

### DIFF
--- a/frontend/src/components/HandbookOverlay/Pages/HandbookSystemRole.tsx
+++ b/frontend/src/components/HandbookOverlay/Pages/HandbookSystemRole.tsx
@@ -1,4 +1,4 @@
-import { LEVEL_NAMES, LevelSystemRole } from '@src/models/level';
+import { LevelSystemRole } from '@src/models/level';
 
 import './HandbookPage.css';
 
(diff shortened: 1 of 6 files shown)


frontend/src/components/HandbookOverlay/Pages/HandbookSystemRole.tsx
```diff
@@ -1,12 +1,12 @@
...
+import { LevelSystemRole } from '@src/models/level';
 
 import './HandbookPage.css';
 
 function HandbookSystemRole({
 	numCompletedLevels,
 	systemRoles,
 }: {
-	numCompletedLevels: LEVEL_NAMES;
+	numCompletedLevels: number;
```
chriswilty, line 9: Yeah this makes more sense 👍

chriswilty (approved):
Beautiful 💎

PR #844, opened 2024-02-23
Stop sending full defences on low levels

pmarsh-scottlogic:
## Description

Level 1 and and 2 don't have defences. So they don't need defences stored in the backend session, nor do they need to be passed to the frontend. So I've got rid of 'em

## Notes

- For level 1 and 2, we store undefined in the defences part of the session rather that a big defences object.
- While I was there, I made the endpoint validation a bit more stringent

## Checklist

Have you done the following?

- [ ] Linked the relevant Issue - no issue
- [ ] Added tests


backend/src/controller/defenceController.ts
```diff
@@ -28,33 +28,74 @@ function handleDefenceActivation(req: DefenceActivateRequest, res: Response) {
 	const defenceId = req.body.defenceId;
 	const level = req.body.level;
```
chriswilty, line 29: `const { defenceId, level } = req.body;`

backend/src/server.ts
```diff
@@ -24,7 +24,8 @@ app.listen(port, () => {
 			console.debug('Document vector store initialized');
 		})
 		.catch((err) => {
-			throw new Error(`Error initializing document vectors: ${err}`);
+			//throw new Error(`Error initializing document vectors: ${err}`);
+			console.log('ERROR SWALLOWED :D', err);
```
chriswilty, line 28: Erm... No?

backend/src/controller/defenceController.ts
```diff
@@ -28,33 +28,74 @@ function handleDefenceActivation(req: DefenceActivateRequest, res: Response) {
...
 		);
-		res.status(200).send();
-	} else {
-		res.status(400).send();
+		return;
 	}
+
+	if (req.session.levelState[3].defences === undefined) return; // this dumb type guard is needed, because we are accessing the array via enum, and so typescript doesn't account for the above type guards. This will be solved with #842. The below array indices won't need to be hardcoded either.
+
+	req.session.levelState[3].defences = activateDefence(
```
chriswilty, line 57: I am confused. We have undefined defences on LEVEL_1 and LEVEL_2, but not on LEVEL_3 or SANDBOX, right? But here you are always activating a SANDBOX defence, never a LEVEL_3 defence.

Same for the other handlers, although you cannot configure or reset defences on LEVEL_3, only in SANDBOX.

chriswilty (changes requested):
Not sure if I'm reading it correctly, but looks like there's a logical error here for LEVEL_3 defences. If your test coverage was better, those would be able to tell me if my thinking was correct or not 😉

backend/src/controller/defenceController.ts
pmarsh-scottlogic: oh you're right. This is about to be fixed with the levels not being an enum any more

pmarsh-scottlogic:
Right. I've done a bunch more stuff
- removed the weird type guarding I was doing before, and made it so that levels 3 and sandbox can be configured.
- Revamped the validation across the defenceController endpoints because it was a bit naff.
- Altered `handleResetSingleDefence` and related code (backend and frontend).
  - renamed to  `handleResetDefenceConfigItem`
  - renamed the stuff that just said `configId` to `configItemId`
  - Now takes a level parameter to be in line with the other endpoints, even if it's technically not necessary. I think consistency is more important. Don't know if that will be a popular decision.
- adds full unit testing for defenceController
- adds integration testing for defenceController, which make sure that the right levels can and can't be configured. The reason these are integration tests and not unit tests is because a level's configurability is determined by whether its `defences` property is initially undefined or not.

backend/src/controller/defenceController.ts
```diff
@@ -43,7 +42,9 @@ function handleDefenceActivation(req: DefenceActivateRequest, res: Response) {
 		return;
 	}
 
-	if (req.session.levelState[level].defences === undefined) {
+	const currentDefences = req.session.levelState[level].defences;
+
+	if (currentDefences === undefined) {
```
chriswilty, line 47: `if (!currentDefences)` is a little more robust, as it can handle null or undefined.

Same further down.

backend/src/controller/defenceController.ts
```diff
@@ -89,23 +96,40 @@ function handleDefenceDeactivation(req: DefenceActivateRequest, res: Response) {
...
 }
 
 function handleConfigureDefence(req: DefenceConfigureRequest, res: Response) {
-	// id of the defence
-	const defenceId = req.body.defenceId;
-	const config = req.body.config;
-	const level = req.body.level;
+	const { defenceId, level, config } = req.body;
+
+	if (defenceId === undefined) {
```
chriswilty, line 116: `if (!defenceId)` also guards against empty string or null.

backend/src/controller/defenceController.ts
```diff
@@ -89,23 +96,40 @@ function handleDefenceDeactivation(req: DefenceActivateRequest, res: Response) {
...
+		sendErrorResponse(res, 400, 'Missing defenceId');
+		return;
+	}
+
+	if (level === undefined) {
```
chriswilty, line 121: `if (!Number.isFinite(level))` will guard against undefined, null or NaN.

backend/src/controller/defenceController.ts
```diff
@@ -89,23 +96,40 @@ function handleDefenceDeactivation(req: DefenceActivateRequest, res: Response) {
...
+	}
+
+	if (level < LEVEL_NAMES.LEVEL_1 || level > LEVEL_NAMES.SANDBOX) {
+		sendErrorResponse(res, 400, 'Invalid level');
+		return;
+	}
 
-	if (!defenceId || !config || level === undefined) {
-		sendErrorResponse(res, 400, 'Missing defenceId, config or level');
+	if (config === undefined) {
```
chriswilty, line 131: `if (!config)`

backend/src/controller/defenceController.ts
```diff
@@ -114,7 +138,9 @@ function handleConfigureDefence(req: DefenceConfigureRequest, res: Response) {
 		return;
 	}
 
-	if (req.session.levelState[level].defences === undefined) {
+	const currentDefences = req.session.levelState[level].defences;
+
+	if (currentDefences === undefined) {
```
chriswilty, line 143: `if (!currentDefences)`

backend/test/integration/defenceController.test.ts
```diff
@@ -0,0 +1,312 @@
...
+import { DefenceConfigItemResetRequest } from '@src/models/api/DefenceConfigResetRequest';
+import { DefenceConfigureRequest } from '@src/models/api/DefenceConfigureRequest';
+import { DefenceStatusRequest } from '@src/models/api/DefenceStatusRequest';
+import { ChatModel } from '@src/models/chat';
+import { ChatMessage } from '@src/models/chatMessage';
+import { DEFENCE_ID, Defence } from '@src/models/defence';
+import { EmailInfo } from '@src/models/email';
+import { LEVEL_NAMES, getInitialLevelStates } from '@src/models/level';
+
+declare module 'express-session' {
```
chriswilty, line 21: This will need removing, once your other PR is in and this one rebased.

backend/test/integration/defenceController.test.ts
```diff
@@ -0,0 +1,312 @@
...
+
+			expect(res.status).toHaveBeenCalledWith(400);
+			expect(res.send).toHaveBeenCalledWith(
+				'You cannot reset defence config items on this level, because it uses the default defences'
+			);
+		});
+
+		test(`GIVEN level ${
+			level + 1
+		} WHEN attempt to get defence status THEN shoult return undefined`, () => {
```
chriswilty, line 140: Typo: _should_

Is this logic correct though? Do we not expect a BadRequest response?

backend/test/integration/defenceController.test.ts
```diff
@@ -0,0 +1,312 @@
...
+			const res = responseMock();
+
+			handleDefenceDeactivation(req, res);
+
+			expect(res.status).toHaveBeenCalledWith(200);
+			expect(res.send).toHaveBeenCalled();
+			expect(req.session.levelState).toEqual(getInitialLevelStates());
+		});
+
+		test(`GIVEN level ${level} WHEN attempt to configure a defence THEN defence is configured`, () => {
```
chriswilty, line 224: I don't think so - we can only configure defences in SANDBOX. Same for reset. We can activate and deactivate in Level 3 though.

If this test passes, then I'd say your implementation logic is incorrect.

chriswilty (changes requested):
Lovely work, lots going on here! All those defence controller unit tests made my eyes glaze over 😅

backend/src/controller/defenceController.ts
pmarsh-scottlogic: I've changed them to this `if (!Number.isFinite(level) || level === undefined)` for the sake of type narrowing

backend/test/integration/defenceController.test.ts
pmarsh-scottlogic: Irrelevant now, the get defences endpoint no longer exists

backend/test/integration/defenceController.test.ts
pmarsh-scottlogic: I was not going to bother locking down the API against configuring defences for level 3 for the sake of simplicity. I'll change that

pmarsh-scottlogic: done

pmarsh-scottlogic:
I've restructured the tests in integration/defenceController.test. They now use the neat `test.each()` syntax that I saw in your other PR, and they're structured by controller method. The tests themselves haven't seen any change, apart from configure defence and reset config item, which now make sure that an error response is returned if on levels 1, 2 or 3.

backend/src/controller/chatController.ts
```diff
@@ -240,14 +241,15 @@ async function handleChatToGPT(req: OpenAiChatRequest, res: Response) {
 			? req.session.chatModel
 			: defaultChatModel;
 
+	const defences =
```
chriswilty, line 244: I'm wondering why we want to define defences here (other than the system role) for levels 1 and 2, which leads me to wonder why handleChatWithoutDefenceDetection needs to receive the defences at all...

If you follow the passing of defences through handleChatWithoutDefenceDetection and beyond, you end up at `handleAskQuestionFunction` in src/openai.ts, which simply checks if the Q&A LLM defence is active. If we don't have _any_ defences, then surely we can assume it's not active. Plus, it feels overly complex to be passing all the defences through so many calls, when all we need is the QA_LLM defence.

backend/src/controller/defenceController.ts
```diff
@@ -24,46 +24,99 @@ function configValueExceedsCharacterLimit(config: DefenceConfigItem[]) {
...
+
+	if (!Number.isFinite(level) || level === undefined) {
+		sendErrorResponse(res, 400, 'Missing level');
+		return;
+	}
+
+	if (level < LEVEL_NAMES.LEVEL_1 || level > LEVEL_NAMES.SANDBOX) {
+		sendErrorResponse(res, 400, 'Invalid level');
+		return;
+	}
```
chriswilty, line 42: We appear to have these same three checks in four different handler functions now. Any way we can extract them?

frontend/src/service/defenceService.ts
```diff
@@ -63,60 +65,61 @@ async function configureDefence(
...
 	// config is a number greater than zero
-	return !isNaN(Number(config)) && Number(config) > 0;
+	return !isNaN(Number(value)) && Number(value) > 0;
 }
 
-function validateNonEmptyStringConfig(config: string) {
+function validateNonEmptyStringConfig(value: string) {
 	// config is non empty string and is not a number
-	return config !== '' && !Number(config);
+	return value !== '' && !Number(value);
```
chriswilty, line 90: Unrelated to this issue, but does this strike you as odd? We are validating a string is non-empty, but we are also disallowing numeric-only content such as "123456". And yet something like "123 456" or "123,456" or "123_456" _is_ allowed cos the result will be `NaN`.

I'll add it to my todo list to investigate.

frontend/src/service/defenceService.ts
```diff
@@ -63,60 +65,61 @@ async function configureDefence(
...
 }
 
-function validateFilterConfig(config: string) {
+function validateFilterConfig(value: string) {
 	// config is not a list of empty commas
 	const commaPattern = /^,*,*$/;
-	return config === '' || !commaPattern.test(config);
+	return value === '' || !commaPattern.test(value);
```
chriswilty, line 96: Ha, another unrelated thing!

This "comma pattern" will actually match the empty string. Pretty sure that's not what we wanted, seeing as we needed to specifically allow empty string to begin with.

We probably want `/^,+$/`, or better, we split the string on commas, trim() each entry, remove falsy values, and then check if the resulting list is empty or not.

Another to add to my todo list 😉

chriswilty (commented):
A few minor thoughts, but this looks good now 👍

backend/test/unit/controller/levelController.test.ts
```diff
@@ -0,0 +1,119 @@
...
+	mockSend.mockClear();
+});
+
+[
+	LEVEL_NAMES.LEVEL_1,
+	LEVEL_NAMES.LEVEL_2,
+	LEVEL_NAMES.LEVEL_3,
+	LEVEL_NAMES.SANDBOX,
+].forEach((level) => {
+	test(`GIVEN level ${
```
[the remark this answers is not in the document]
pmarsh-scottlogic: oh yes, should've been more thorough

backend/test/unit/controller/startController.test.ts
```diff
@@ -0,0 +1,126 @@
...
+				],
+				systemRoles: [],
+			},
+		} as unknown as StartGetRequest;
+		const res = responseMock();
+
+		handleStart(req, res);
+
+		expect(mockSend).toHaveBeenCalledWith({
+			emails: `level ${level + 1} emails`,
```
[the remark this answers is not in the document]
pmarsh-scottlogic: Not certain what you're referring to. I'm going to assume it's the fact that we're setting and returning string values in our mocked session rather than the types we have set elsewhere.

Isn't it a good thing for tests to be able to ignore the complex types of the real thing, so that the tests can be kept as simple as possible?

Anyway, this is only allowed because of the type casting: `const req = {...} as unknown as StartGetRequest`. We're explicitly and intentionally going over the type system.

backend/src/controller/defenceController.ts
pmarsh-scottlogic: ah yes, I added a method to validate level in my other PR :) just need to switch it here

pmarsh-scottlogic: done

frontend/src/service/defenceService.ts
pmarsh-scottlogic: good spot. Would be a nice thing to fix. Though you could argue that a nonsense numerical config (say, system prompt) is still valid even if it won't do much good.

frontend/src/service/defenceService.ts
pmarsh-scottlogic: that's easy enough to do here

frontend/src/service/defenceService.ts
pmarsh-scottlogic: I was about to change this too, and then I thought I could add tests, but I stopped myself before this PR bloats even more

backend/src/controller/chatController.ts
pmarsh-scottlogic: what do you suggest? pass undefined through the calls instead? Make new methods for the whole call chain for getting a response without defences? I vote the former

backend/test/unit/controller/startController.test.ts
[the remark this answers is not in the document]
chriswilty: Yes that's what I mean. We are verifying that a function was called with something it will (and can) never be called with in the actual implementation. Just looked odd that's all, doesn't really matter as we're only using those "somethings" to verify we were using the correct level.

chriswilty (commented):
Yep looks good, just the question of what to do with the defences param, which I've answered above.

backend/src/controller/chatController.ts
pmarsh-scottlogic: The answer is: to get something from the session, we need access to the `req` object

chriswilty: Of course, silly me! As discussed, let's try passing just the Q&A LLM defence down the call chain, but also have it optional so we don't need it for levels 1 and 2.

backend/src/models/defence.ts
```diff
@@ -23,11 +23,24 @@ type DefenceConfigItem = {
 	value: string;
 };
 
-type Defence = {
-	id: DEFENCE_ID;
+type QaLlmDefence = {
+	id: DEFENCE_ID.QA_LLM;
 	config: DefenceConfigItem[];
 	isActive: boolean;
 };
```
chriswilty, line 30: To avoid repetition, you can probably do something like this:

```
type QaLlmDefence = Defence & {
  id: DEFENCE_ID.QA_LLM
};
```

backend/src/openai.ts
```diff
@@ -141,14 +141,14 @@ function isChatGptFunction(functionName: string) {
...
+	qaLlmDefence?: QaLlmDefence
 ) {
 	if (functionCallArgs) {
 		const params = JSON.parse(functionCallArgs) as FunctionAskQuestionParams;
 		console.debug(`Asking question: ${params.question}`);
 		// if asking a question, call the queryDocuments
-		const configQAPrompt = isDefenceActive(DEFENCE_ID.QA_LLM, defences)
-			? getQAPromptFromConfig(defences)
+		const configQAPrompt = qaLlmDefence?.isActive
+			? getQAPromptFromConfig([qaLlmDefence])
```
chriswilty, line 151: If you don't change Defence to that union type, you might need to do `[qaLlmDefence as Defence]` here.

pmarsh-scottlogic: seems to work fine before and after changing defence from that union type

chriswilty (approved):
Lovely ⚡

PR #845, opened 2024-02-27
821 Spike Testing Script

### LINKED ISSUE #821: K6 Spike Testing
Adding in a spike test script and equivalent tests within the script

A spike test verifies whether the system survives and performs under sudden and massive rushes of utilization. Spike testing increases to extremely high loads in a very short or non-existent ramp-up time. Usually, it has no plateau period or is very brief, as real users generally do not stick around doing extra steps in these situations. In the same way, the ramp-down is very fast or non-existent, letting the process iterate only once.

Image

Example script code for spike testing:
```
import http from 'k6/http';
import {sleep} from 'k6';

export const options = {
  // Key configurations for spike in this section
  stages: [
    { duration: '2m', target: 2000 }, // fast ramp-up to a high point
    // No plateau
    { duration: '1m', target: 0 }, // quick ramp-down to 0 users
  ],
};

export default () => {
  const urlRes = http.get('https://test-api.k6.io');
  sleep(1);
  // MORE STEPS
  // Add only the processes that will be on high demand
  // Step1
  // Step2
  // etc.
};
```

kclark-scottlogic:
## Description

Resolves #821  

## Checklist

Have you done the following?


k6/spike.js
```diff
@@ -0,0 +1,49 @@
...
+const cookieName = 'prompt-injection.sid';
+const vuCookieJar = (() => {
+	const cookieJars = {};
+	return {
+		get: (id) => cookieJars[id],
+		set: (id, jar) => (cookieJars[id] = jar),
+	};
+})();
+
+export default () => {
```
chriswilty, line 23: Is this the same function as in load.js? If so, extract the function into its own module and use it in both test files.

k6/spike.js
```diff
@@ -0,0 +1,49 @@
+import { check, sleep } from 'k6';
+import exec from 'k6/execution';
+import http from 'k6/http';
+
+export const options = {
+	// Key configurations for spike in this section
+	stages: [
+		{ duration: '2m', target: 2000 }, // fast ramp-up to a high point
```
chriswilty, line 8: 😰 That's a lot of users!

Seems unrealistic, even if our site is under attack - because then AWS firewall will kick in to rate-limit requests. I'd suggest starting with say 500, and seeing if it breaks. Even 500 seems a lot, but at least it's testing an _unexpectedly_ high load.

chriswilty (changes requested):

chriswilty:
@kclark-scottlogic Can you remember to modify or remove the parts of the PR template you don't need?

k6/spike.js
kclark-scottlogic: Will bring down, I was playing with numbers yesterday and found local breaks around 5000 ish if the uptake is 2m

chriswilty: Well, I'm fine with you leaving this at 2000, though of course when we want to run tests against remote, we could _possibly_ hit AWS rate-limiting... Let's see!

k6/load.js
```diff
@@ -13,40 +11,6 @@ export const options = {
...
-	check(response, {
-		'response code was 200': (response) => response.status === 200,
-		'cookie was preserved': (response) =>
-			response.cookies[cookieName].length === 1 &&
-			response.cookies[cookieName][0].value === expectedCookie,
-	});
-
-	sleep(1);
-};
+export default async function () {
```
chriswilty, line 14: Why `async` ?

I think this can probably just be `export default requestFunction;`

k6/spike.js
```diff
@@ -0,0 +1,13 @@
...
+
+export const options = {
+	// Key configurations for spike in this section
+	stages: [
+		{ duration: '2ms', target: 2000 }, // fast ramp-up to a high point
+		{ duration: '1m', target: 0 }, // quick ramp-down to 0 users
+	],
+};
+
+export default async function () {
```
chriswilty, line 11: Same here.

chriswilty (changes requested):
Looking good, just a couple of suggestions.

k6/load.js
```diff
@@ -11,6 +11,6 @@ export const options = {
 	},
 };
 
-export default async function () {
-	requestFunction();
-}
+export default () => {
+	postWithSessionUpdate();
+};
```
chriswilty, line 16: Can this not just be `export default postWithSessionUpdate`?

PR #849, opened 2024-02-29
Split github workflows so jobs only run on related code changes

### LINKED ISSUE #847: Github actions should only run on modules with code changes
I note that all our github code checks are currently defined in the same workflow file, meaning they all run regardless of code changes. This also costs us more money.

We can easily split our checks into modules - frontend, backend, k6, and soon also cloud, and configure them only to run on a PR if it includes code changes in that module. This will not only speed up PR checks, it'll also be cheaper.

chriswilty:
## Description

Splits our github Pull Request workflow into one per module, so that we can run jobs only when code changes.

Resolves #847 

## Notes

All these improvements will speed up our PR checks, and also potentially reduce our costs:

- Sparse checkout, only fetch the code needed to run a job
- Run job only on related code changes, e.g. no need to run frontend code checks if PR only has backend code changes
- Run build/test in parallel with lint/format checks

## Concerns

- These checks will not run until they are in dev branch, so they won't run on this PR. I don't think we can even trigger them manually yet, as if I recall, they don't appear in the github Actions tab until they are in dev. So, we won't be able to test the new jobs until we start opening subsequent PRs.... 🍀 

## Checklist

Have you done the following?

- [ ] Added tests


dhinrichs-scottlogic (approved):
Looks good :) 
If you want, you could mention load testing in the K6 name somewhere, so that juniors know what to google for

PR #854, opened 2024-03-04
853: Service availability in AWS

### LINKED ISSUE #853: Scale service at start and end of day
I've noticed that AWS costs are about $3.50 per day when the fargate service and load balancer are up and running.

We can reduce these costs by bringing the service down at end of day, then up again at start of next, plus only having the service available mon-fri. At this stage it's unclear how that will affect availability of the load balancer, so we might need to bring that up/down as well, unless Fargate manages that for us.

In the UI, we might need to show some sort of "webapp is offline" page when the service is down, instead of showing the full app with a "could not reach service" message in the chat panel. Need to investigate this, but in any case it might be better as a separate issue.

chriswilty:
## Description

It came to my attention that the fargate service and associated infrastructure is costing us more than $3 a day. We can decrease that by around 70% if we bring the service down overnight, and not have the application up over weekends.

Currently there is no _trivial_ way to achieve this using CDK, although there are some alpha constructs in development which we can use, together with a lambda function, to update the service to flip "desired task count" from zero to 1 in the morning and back to zero in the evening.

Resolves #853 

## Screenshots

image

## Concerns

- We might want to show a different UI page outside of uptime hours, such as a generic "downtime / maintenance" message. I will assess how this looks once it's live in AWS, and spawn another issue if needed.

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/cloud/package.json
+++ b/cloud/package.json
@@ -19,15 +19,18 @@
 		"@types/source-map-support": "^0.5.10",
 		"@typescript-eslint/eslint-plugin": "^6.21.0",
 		"@typescript-eslint/parser": "^6.21.0",
+		"aws-cdk": "^2.131.0",
 		"concurrently": "^8.2.2",
 		"eslint": "^8.56.0",
-		"aws-cdk": "^2.127.0",
 		"prettier": "^3.2.5",
 		"ts-node": "^10.9.2",
 		"typescript": "^5.3.3"
 	},
 	"dependencies": {
-		"aws-cdk-lib": "^2.127.0",
+		"@aws-cdk/aws-scheduler-alpha": "^2.131.0-alpha.0",
+		"@aws-cdk/aws-scheduler-targets-alpha": "^2.131.0-alpha.0",
+		"aws-cdk-lib": "^2.131.0",
+		"@aws-sdk/client-ecs": "^3.525.0",
 		"constructs": "^10.3.0",
 		"dotenv": "^16.4.4",
 		"source-map-support": "^0.5.21"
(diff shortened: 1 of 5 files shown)


cloud/package.json
```diff
@@ -19,15 +19,18 @@
 		"@types/source-map-support": "^0.5.10",
 		"@typescript-eslint/eslint-plugin": "^6.21.0",
 		"@typescript-eslint/parser": "^6.21.0",
+		"aws-cdk": "^2.131.0",
```
pmarsh-scottlogic, line 22: what makes this a devdependency and not a standard dependency?

chriswilty: This is the CLI for running commands, such as "cdk synth", so it's not part of any built artifacts or build process. It's like using jest to run tests, or typescript to compile the code.

In contrast, aws-cdk-lib is used in our scripts that are run to produce the CloudFormation templates, which are then deployed to AWS.

chriswilty:
@pmarsh-scottlogic Understood! It's more complex than I would've liked, but the idea is that we schedule a pair of events, one in morning and one in evening, and when they run, the scheduler turns them into an "action" to run a lambda function. That lambda function has one job to do: adjust the number of tasks/containers the service is running: to 1 in the morning, 0 in the evening.

We need to adjust the service definition itself, because simply auto-scaling the running service to zero will cause Fargate to check its task definition and see that it needs at least one container running, so it immediately brings one up again. This is Fargate's job - it manages container deployment for us, optimising according to current traffic and our defined limits. But our "desired count" is the baseline to which our number of containers will return under "normal" load, and that desired count cannot be zero, as that would make no sense. Hence why this is a touch more complex than I would've liked, and hence the need for a lambda function to adjust the service definition.

PR #855, opened 2024-03-05
824: bugfix for level queryparam

chriswilty:
## Description

Fixes a bug in which the level param was not parsed into a number before checking, so the level endpoint always failed with 400 response.

Related to #824 and #842 

## Checklist

Have you done the following?

- [x] Added tests

### DIFF
--- a/backend/src/models/api/StartGetRequest.ts
+++ b/backend/src/models/api/StartGetRequest.ts
@@ -5,15 +5,20 @@ import { Defence } from '@src/models/defence';
 import { EmailInfo } from '@src/models/email';
 import { LEVEL_NAMES } from '@src/models/level';
 
+export type StartResponse = {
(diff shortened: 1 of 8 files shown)


backend/src/models/api/StartGetRequest.ts
```diff
@@ -5,15 +5,20 @@ import { Defence } from '@src/models/defence';
...
+};
+
 export type StartGetRequest = Request<
 	never,
-	{
-		emails: EmailInfo[];
-		chatHistory: ChatMessage[];
-		defences: Defence[];
-		availableModels: string[];
-		systemRoles: string[];
```
chriswilty, line 15: Note this type was incorrect, now fixed.

pmarsh-scottlogic (approved):
does the trick

Issue #754, opened 2024-01-10
Update focus style for Prompt Enclosure radio buttons


pmarsh-scottlogic:
# Feature Request

## Description

Have the radio button focus match the nice thick blue indicators that are on the other elements!

## Screenshots

Here's what we have
![Image](https://github.com/ScottLogic/prompt-injection/assets/118171430/b53f73d5-0c95-4309-8988-9a8947f1fadf)

We should have something more like this
![Image](https://github.com/ScottLogic/prompt-injection/assets/118171430/cdf5cb91-8cb7-41f9-bb50-feec462c699c)

## Acceptance criteria

**WHEN** A radio button in the Prompt Enclosure defence settings is focused
**THEN** It should have a nice thick blue focus indicator, similar in appearance to the focus indicators on other elements

chriswilty:
I noted that at zoom levels above 100%, the focus ring is a little cropped, on Firefox browser. Possibly needs a little more padding.

PR #857, opened 2024-03-06
fixing radio button centering

AAloshine-scottlogic:
## Description
added the same styling of the toggle switches to the radio buttons for focusing and hovering, recentered the selected radio button dot

## Screenshots
before
image

after
image
image

note: maybe text align to the left so the overflow text on the next line isn't as close to the highlight? or i could increase the margin from 0.5 to 0.6rems to give the radio button a little more space?

## Checklist

Have you done the following?

https://github.com/ScottLogic/prompt-injection/issues/754
- [ ] Added tests

### DIFF
--- a/frontend/src/components/DefenceBox/DefenceConfigurationRadioButton.css
+++ b/frontend/src/components/DefenceBox/DefenceConfigurationRadioButton.css
@@ -2,15 +2,12 @@
 	position: relative;
 	display: flex;
 	cursor: pointer;
+	gap: 0.75rem;
 }
 
 .defence-radio-button input {
 	margin-bottom: 0.5rem;
 	opacity: 0;
-}
-
-.defence-radio-button span {
-	margin-left: 1rem;
 }
 
 /* checkmark outer circle */
--- a/frontend/src/components/DefenceBox/DefenceConfigurationRadioButton.css
+++ b/frontend/src/components/DefenceBox/DefenceConfigurationRadioButton.css
@@ -1,9 +1,9 @@
 .defence-radio-button {
 	position: relative;
 	display: inline-block;
+	display: flex;
 	margin: 0.1rem;
 	cursor: pointer;
-	display: flex;
 }
 
 .defence-radio-button input {
--- a/frontend/src/components/DefenceBox/DefenceConfigurationRadioButton.css
+++ b/frontend/src/components/DefenceBox/DefenceConfigurationRadioButton.css
@@ -1,10 +1,12 @@
 .defence-radio-button {
 	position: relative;
 	display: inline-block;
+	margin: 0.1rem;
 	cursor: pointer;
 }
 
 .defence-radio-button input {
+	margin-bottom: 0.5rem;
 	opacity: 0;
 }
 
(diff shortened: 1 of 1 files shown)


chriswilty:
@AAloshine-scottlogic I'd say we put the radiobuttons and labels into some sort of container - flex or grid - so that when the text wraps, it looks like this:

radiobuttons

frontend/src/components/DefenceBox/DefenceConfigurationRadioButton.css
```diff
@@ -28,8 +31,8 @@
 .defence-radio-button input + span::after {
 	content: '';
 	position: absolute;
-	top: 0.3125rem;
-	left: 0.3125rem;
+	top: 0.34rem;
+	left: 0.34rem;
```
pmarsh-scottlogic, line 35: This change has caused the little "pupil" of the bottom two radio buttons to be off-centre. What was the intention of the change?

image

pmarsh-scottlogic (changes requested):
Just one wee problem from me. Otherwise, solid.

frontend/src/components/DefenceBox/DefenceConfigurationRadioButton.css
AAloshine-scottlogic: the original had the button off centered (if you look at the before picture)

pmarsh-scottlogic: weird. For me it's on centre on my machine (big 1080p monitor, full screen, no zoom), and then off-centre after the change

AAloshine-scottlogic: well its not in the scope of this ticket so i'll just revert it

frontend/src/components/DefenceBox/DefenceConfigurationRadioButton.css
```diff
@@ -1,10 +1,13 @@
 .defence-radio-button {
 	position: relative;
 	display: inline-block;
+	display: flex;
+	margin: 0.1rem;
```
chriswilty, line 5: Remove this, seems unnecessary. We already have `gap: 0.5rem` on the parent element, so if you want more vertical space between radios, bump that up a little.

Also, I'd prefer `gap: 0.75rem` on this class, instead of margin-left on `.defence-radio-button span` below.

frontend/src/components/DefenceBox/DefenceConfigurationRadioButton.css
```diff
@@ -1,10 +1,13 @@
 .defence-radio-button {
 	position: relative;
 	display: inline-block;
+	display: flex;
```
chriswilty, line 4: Remove the `display: inline-block` here.

chriswilty (changes requested):
Looks good, very minor change suggestions only.

What Peter said about the off-centred dot, is what I was hinting at with my "check at various zoom levels" because for whatever reason, this radiobutton styling is not consistent across different browsers or zoom levels.

I am minded to make it all a bit simpler anyway... But that's for another day!

frontend/src/components/DefenceBox/DefenceConfigurationRadioButton.css
chriswilty: I think we can solve this with a simpler css structure, but that's for another day. I've seen perfectly good radiobuttons with custom styling that only have a `::before` pseudo-element, not both before and after, which is what seems to be throwing alignment off here.

chriswilty:
@pmarsh-scottlogic Just waiting on a re-review from you, then this one can go in.

pmarsh-scottlogic (approved):
fab

PR #858, opened 2024-03-07
839 optionally return chat model in start controller and level controller

pmarsh-scottlogic:
⚠NOTE: I've put the wrong ticket number on the branch name.

## Description

The final PR for #824  !
This one is to stop the getModel calls. We now return the current configured model in getStart or getLevel (but only for sandbox).

### Before

- Refreshing on sandbox (preview mode, so only single render):
  image

- changing level to sandbox (preview mode, so only single render):
  image

### After

- Refreshing on sandbox (preview mode, so only single render):
  image

- changing level to sandbox (preview mode, so only single render):
  image

## Notes

Here's the whole situation because it's a bit convoluted:
- In the backend session we store `chatModel`. This records the id of the model that we want to chat to as well as the configurable settings (_temperature, top p,_ etc). But we use the default `chatModel` rather than this configured version for all levels apart from sandbox. So really what this value represents is the `chatModel` used in _sandbox_ (despite the name and it not being part of the `levelState`). I think this is confusing, but I tried to change it and it snowballed into a bit of a mess, so 'm calling it a problem for another day.
- Anyhoo, we need to get the configured chatModel whenever we load up or switch to the sandbox. So I've attached a chatModel attribute to the `StartResponse` and `LoadLevelResponse`. But because this is only relevant for sandbox, on all other levels this value is undefined.
- Meanwhile in the frontend, in `MainComponent` we consume these response objects. We set the `chatModel` useState value, and then drill that down all the way to the `ModelSelection` and `ModelConfiguration` components, which show up in the control panel in sandbox.
- I've opted to sync the frontend's copy of the `chatModel` with the backend's one, i.e. it will only store the configured chatModel used in the sandbox.

## Concerns

- I'm sure there are nicer, less confusing ways to organise all the data, but i don't think it's this ticket's job to do that. (read: I tried to reorganise it but it grew arms and legs, so I'm choosing to revert and then ignore the problem)
- I wonder if I should rename `chatModel` across the front and back end to something like `sandboxChatModel` just to make it really clear what it's there for.

## Checklist

Have you done the following?

- [ ] Added tests


backend/src/models/api/LevelGetRequest.ts
```diff
@@ -11,6 +12,7 @@ export type LevelGetRequest = Request<
 		emails: EmailInfo[];
 		chatHistory: ChatMessage[];
 		defences: Defence[];
+		chatModel: ChatModel;
```
chriswilty, line 15: Not that it matters very much, but I assume this chatModel field is actually optional, as we only receive it for Sandbox. I think it would help with clarity to mark it as such.

frontend/src/components/MainComponent/MainComponent.tsx
```diff
@@ -167,6 +180,9 @@ function MainComponent({
 			: setMessages(chatHistory);
 
 		setDefences(defences);
+
+		// we will only update the chatModel if it is defined in the backend response. It will only defined for the sandbox level.
+		chatModel && setChatModel(chatModel);
```
chriswilty, line 185: Is this correct? It means if we switch from sandbox to any other level, we will still have sandbox chatModel in state. Not only is that an unnecessary use of memory, but it's also an inconsistency.

frontend/src/components/MainComponent/MainComponent.tsx
```diff
@@ -283,6 +299,16 @@ function MainComponent({
 		setMessages((messages: ChatMessage[]) => [resetMessage, ...messages]);
 	}
 
+	function setChatModelId(modelId: string) {
+		if (!chatModel) {
+			console.error(
+				'You are trying to change the id of the chatModel but it has not been loaded yet'
```
chriswilty, line 305: Not convinced this error message will help a future developer work out what's going on here. Can you come up with something indicating we only expect chat model to be changed on Sandbox level?

frontend/src/components/ModelBox/ModelConfiguration.tsx
```diff
@@ -1,17 +1,24 @@
...
 import './ModelConfiguration.css';
 
 function ModelConfiguration({
+	chatModel,
 	addInfoMessage,
 }: {
+	chatModel?: ChatModel;
 	addInfoMessage: (message: string) => void;
 }) {
+	// can we move the default configurations to a different file?
```
chriswilty, line 21: Either that or just outside of the component itself. Doesn't _need_ to be in a different file.

frontend/src/components/ModelBox/ModelConfiguration.tsx
```diff
@@ -79,25 +86,18 @@ function ModelConfiguration({
...
-					}
-				);
-				setCustomChatModel(newCustomChatModelConfigs);
-			})
-			.catch((err) => {
-				console.log(err);
-			});
-	}, []);
+		if (!chatModel) {
+			// chatModel is undefined if this is the first time that the user has switched to the sandbox level, and the change level request has not yet resolved successfully
```
chriswilty, line 91: Can we wrap this comment line please?

frontend/src/components/ModelBox/ModelConfiguration.tsx
```diff
@@ -79,25 +86,18 @@ function ModelConfiguration({
...
+			return;
+		}
+		const newCustomChatModelConfigs = customChatModelConfigs.map((config) => {
+			const newConfig = { ...config };
+			newConfig.value = chatModel.configuration[config.id];
+			return newConfig;
+		});
```
chriswilty, line 98:
```
const newCustomChatModelConfigs = customChatModelConfigs.map((config) => ({
  ...config,
  value: chatModel.configuration[config.id],
});
```

For what it's worth, this is potentially wasteful because we are generating a brand new config object for each entry in customChatModelConfigs, even if it hasn't changed, which will trigger unnecessary renders. I seem to recall there's an issue to send and receive only the model config that the user has changed; if so, don't worry about it here.

chriswilty (commented):
A few thoughts at this stage, but looks sound 👍

backend/src/models/api/LevelGetRequest.ts
pmarsh-scottlogic: whoops, fixed

frontend/src/components/MainComponent/MainComponent.tsx
pmarsh-scottlogic: I had reasons for this but actually they're bogus. Changed

frontend/src/components/MainComponent/MainComponent.tsx
pmarsh-scottlogic: "available models" arrives under the startResponse, along with chatModel. So if the game is loaded up or refreshed on sandbox, then the available models and chatModel arrive at the same time. However if the user switches from any level to sandbox, then there will be some amount of time (usually negligible) where the chat model dropdown and choose button are rendered, but the chatModel is not yet loaded.

pmarsh-scottlogic: Okay, conditionally rendering the select model bits depending on chatModel being loaded. Wondering if we're doing this then we should also not render the model configuration sliders...
image

frontend/src/components/ModelBox/ModelConfiguration.tsx
pmarsh-scottlogic: moved out of the component

frontend/src/components/ModelBox/ModelConfiguration.tsx
pmarsh-scottlogic: can't find the issue so I'll implement here

pmarsh-scottlogic: There are no extra renders because `chatModel`, which we pass into this component, only changes on refresh of level change. Updating a config slider will not cause `chatModel` to update. I could change that (to keep them in sync), but I don't see it adding any functional value at this point. I want to avoid it because we'd have to drill another thing all the way down the components

chriswilty: Yep ok, React is likely just optimising it based on keys.

frontend/src/components/MainComponent/MainComponent.tsx
chriswilty: > "available models" arrives under the startResponse, along with chatModel. So if the game is loaded up or refreshed on sandbox, then the available models and chatModel arrive at the same time. However if the user switches from any level to sandbox, then there will be some amount of time (usually negligible) where the chat model dropdown and choose button are rendered, but the chatModel is not yet loaded.

Right, that makes sense now 😅 

I'd say if we're going down this route, then we should conditionally render the sliders as you suggest. Can you see how that looks - in the code I mean, check it's not too gnarly?

chriswilty: As discussed in the standup, we'll leave this for a future issue, as it has some (re)design implications.

chriswilty (approved):
Lovely 🏆

PR #864, opened 2024-03-18
793 separate pdf renderer into own chunk

pmarsh-scottlogic:
## Description

does what is says on the tin. I was going to try and lazy load the docviewer as well, but that opens up annoying complications with importing the `DocViewerRenderers` which I didn't want to deal with.

## Checklist

Have you done the following?

- [ ] Added tests


frontend/src/components/DocumentViewer/DocumentViewBox.tsx
```diff
@@ -59,16 +59,18 @@ function DocumentViewBox({ closeOverlay }: { closeOverlay: () => void }) {
...
-						documents={documentMetas}
-						activeDocument={documentMetas[documentIndex]}
-						pluginRenderers={DocViewerRenderers}
-						config={{
-							header: {
-								disableHeader: true,
-							},
-						}}
-					/>
+					<Suspense fallback={<p>loading...</p>}>
```
chriswilty, line 62: Presumably you don't need this if you're no longer lazy loading anything in this component.

frontend/src/components/ChatBox/ChatBox.tsx
```diff
@@ -232,11 +235,13 @@ function ChatBox({
 					</span>
 				</div>
 				<div className="control-buttons">
-					<ExportPDFLink
-						messages={messages}
-						emails={emails}
-						currentLevel={currentLevel}
-					/>
+					<Suspense fallback={<p>loading pdf library</p>}>
```
chriswilty, line 238: Have you checked what this looks like? You should be able to simulate a long loading using a promise and a timeout.

You might be better off using a disabled, no-op button (possibly even a LoadingButton) as your fallback component.

chriswilty (changes requested):
👍 One thing spotted and one suggestion.

frontend/src/components/ChatBox/ChatBox.tsx
pmarsh-scottlogic: Good shout. The text before was invisible!

I've got it looking like this before the real component is loaded
image

Delaying the loading with this code 
```typescript
const ExportPDFLink = lazy(async () => {
	return await later(10000).then(
		() => import('@src/components/ExportChat/ExportPDFLink')
	);
});

function later(delay: number) {
	return new Promise(function (resolve) {
		setTimeout(resolve, delay);
	});
}
```

pmarsh-scottlogic: while I was at it, I updated the hover colour of the export chat and reset level buttons, because it was practically identical to the default background colour.

frontend/src/components/ChatBox/ChatBox.tsx
```diff
@@ -235,7 +236,23 @@ function ChatBox({
 					</span>
 				</div>
 				<div className="control-buttons">
-					<Suspense fallback={<p>loading pdf library</p>}>
+					<Suspense
+						fallback={
+							<ThemedButton
+								className={'chat-button chat-button-disabled'}
+								onClick={() => {
+									console.error('This button is still loading');
```
chriswilty, line 244: I'd be tempted to use a no-op for click, rather than put an error into the console.

`onClick={() => {}}`

However, that depends on what you do about my other comment for this fallback.

frontend/src/components/ThemedButtons/ChatButton.css
```diff
@@ -9,8 +9,13 @@
...
 	font-weight: 700;
 	font-size: 1rem;
+	width: 100%;
 }
 
 .chat-button:hover {
 	background-color: var(--chat-button-background-colour-hover);
 }
+
+.chat-button-disabled:hover {
```
chriswilty, line 19: You don't need an extra class for this, can use an attribute selector:

`.chat-button[aria-disabled='true']:hover {`

frontend/src/components/ChatBox/ChatBox.tsx
```diff
@@ -235,7 +236,23 @@ function ChatBox({
...
+								}}
+								ariaDisabled={true}
+								tooltip={{
+									id: 'hey',
```
chriswilty, line 248: This id must be unique in the document, so, best to use a meaningful id. In any case, I think we can do without a tooltip because it's expected to be a short-lived loading state. If the module fails to load then we have bigger (network) problems anyway.

Though in that case, it would be useful to keep your console.log statement indicating the button is loading, just in case. But then, make it a log, not an error.

chriswilty (changes requested):
Lovely! Just a couple of suggestions.

frontend/src/Theme.css
```diff
@@ -107,7 +107,7 @@
 	--chat-alerted-text-colour: teal;
 	--chat-triggered-text-colour: var(--error-colour);
 	--chat-button-background-colour: #eee;
-	--chat-button-background-colour-hover: #dedede;
+	--chat-button-background-colour-hover: #cecece;
```
chriswilty, line 110: I assume you've checked contrast is still ok, on hover?

pmarsh-scottlogic: yup, all good

frontend/src/components/ThemedButtons/ChatButton.css
pmarsh-scottlogic: ah, I knew I was missing some sort of trick!

frontend/src/components/ChatBox/ChatBox.tsx
pmarsh-scottlogic: well, the button shouldn't be clickable anyway, so yeah I'll make it an empty method

frontend/src/components/ChatBox/ChatBox.tsx
pmarsh-scottlogic: if we remove the tooltip there is nothing indicating to the user (sighted or on-sighted) why the button is disabled... Which might be okay but I think it's better to have a naff solution than no solution at all

pmarsh-scottlogic: I've got it looking like this 
image

frontend/src/components/ChatBox/ChatBox.tsx
```diff
@@ -232,11 +236,29 @@ function ChatBox({
...
+							<ThemedButton
+								className={'chat-button chat-button-disabled'}
+								// eslint-disable-next-line @typescript-eslint/no-empty-function
+								onClick={() => {}}
+								ariaDisabled={true}
+								tooltip={{
+									id: 'export-chat-tooltip',
+									text: 'This button is still loading. Please wait...',
+								}}
+								tooltipAnchor="bottom-center"
```
chriswilty, line 250: I think if we renamed this prop "tooltipPosition", it would help clarify the API.

Also then you'd need to switch "bottom" and "top".

frontend/src/components/ThemedButtons/ThemedButton.tsx
```diff
@@ -22,7 +26,7 @@ function ThemedButton({
 	ariaLabel,
 	className,
 	tooltip,
-	tooltipAnchor = 'top-center',
+	tooltipPosition = 'top-center',
```
chriswilty, line 29: Think we want the default to be bottom-center now.

chriswilty (changes requested):
One small but important thing, else good to go 👍

frontend/src/components/ThemedButtons/ThemedButton.tsx
pmarsh-scottlogic: Good spot!

chriswilty (approved):
Lovely 👌

PR #865, opened 2024-03-19
827 streamline network calls for reset all progress

pmarsh-scottlogic:
## Description

Combines all network calls involved with resetting all progress into one network call.

## Screenshots

image
One network call when you reset all progress

## Notes

- We now pass the current level into the reset progress endpoint. The endpoint resets the values in the backend session storage and returns the state of the current level, which is then processed by the frontend to update the UI accordingly.
- I've update the endpoint to also reset the chat model configuration. Not in scope of ticket but I think it makes more sense then not resetting it, given all the defence configurations are reset.
- Updates the tests for levelController and startController to account for the changes made in #858, where we only return the chatModel on the sandbox.

## Checklist

Have you done the following?

- [x] Added tests

### DIFF
--- a/frontend/src/components/MainComponent/MainComponent.tsx
+++ b/frontend/src/components/MainComponent/MainComponent.tsx
@@ -322,8 +322,17 @@ function MainComponent({
 	// resets whole game progress and start from level 1 or Sandbox
 	async function resetProgress() {
 		console.log('resetting progress for all levels');
-		await resetService.resetAllLevelProgress(); //yeeeeeeeee
 		resetCompletedLevels();
+
+		const { emails, chatHistory, defences, chatModel } =
+			await resetService.resetAllLevelProgress(currentLevel);
+		processBackendLevelData(
+			currentLevel,
+			emails,
+			chatHistory,
+			defences,
+			chatModel
+		);
 
 		// set as new user so welcome modal shows
 		setIsNewUser(true);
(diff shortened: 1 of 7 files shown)


frontend/src/components/MainComponent/MainComponent.tsx
```diff
@@ -315,6 +319,38 @@ function MainComponent({
...
 	}
 
+	// resets whole game progress and start from level 1 or Sandbox
+	async function resetProgress() {
+		console.log('resetting progress for all levels');
+		resetCompletedLevels();
+
+		const { emails, chatHistory, defences, chatModel } =
+			await resetService.resetAllLevelProgress(currentLevel);
+		processBackendLevelData(
```
chriswilty, line 329: Is this processing necessary if we are on level 2 or 3 and we will be taken back to level 1?

chriswilty (commented):
I look forward to rebasing my bugfix on these changes.... Or vice versa I guess 😅

frontend/src/service/resetService.ts
```diff
@@ -1,15 +1,27 @@
...
+
 import { sendRequest } from './backendService';
+import { getChatMessagesFromDTOResponse } from './chatService';
+import { getDefencesFromDTOs } from './defenceService';
 
 const PATH = 'reset';
 
-async function resetAllLevelProgress(): Promise<boolean> {
-	const response = await sendRequest(PATH, {
+async function resetAllLevelProgress(level: number) {
```
chriswilty, line 9: I wonder if we might change the name of this function for clarity, so it's not confused with functionality to reset a single level. Maybe just "resetAllProgress"?

chriswilty (approved):
Beautiful 💎

PR #866, opened 2024-03-19
Cleanup dialog opening/closing, remove click away to close behaviour

### LINKED ISSUE #862: Occasionally the modal dialogs can no longer be opened
# Bug report

## Description

Occasionally, the application gets itself into a state in which modal dialogs no longer open. This is _really difficult_ to reproduce as we have yet to spot any pattern; this in itself indicates it might be a timing / race-condition issue.

It is not to do with the React double-render in dev/strict mode, as Kira has also spotted this in Vite preview mode.

Some "close dialog" events trigger other dialogs to open, which seems a likely candidate for this bug. In particular, there is complexity _and redundancy_ in the process to open the MissionInformation dialog, as it is triggered when
- OverlayWelcome dialog is closed by selecting a level (level1 or sandbox)
- LevelsComplete dialog is closed by clicking "go to sandbox"
- currentLevel (local state in App.tsx) changes via a call to setCurrentLevel

Some of these calls are duplicated, for example:
- "go to sandbox" from LevelsComplete causes a double call of closeOverlay
- Both "go to sandbox" from LevelsComplete dialog and selecting a level from OverlayWelcome dialog cause a double call to open the MissionInformation dialog.

These are problematic, because it means we have two calls to close the dialog, and one or two calls to open the MissionInformation dialog, in quick succession. If these occur out of intended order, that might cause the app to get into a corrupt state. It's worth noting that the API docs for `HTMLDialogElement.showModal()` state an error is thrown if the dialog is already in the "open" state, but our app is not catching that error. It is not clear from the docs what might happen in that case, but it seems entirely plausible that the dialog simply stops functioning.

In any case, we can start by adding some debug logging statements whenever a dialog is opened or closed, to see if any events fire in an unexpected order. We could then try to reproduce the bug, and then remove the redundancy in function calls and see if we can still reproduce the bug.

## Reproduction steps

As noted above, this is proving difficult to reproduce, but start by repeatedly switching between sandbox and level1 from the Welcome dialog.

## Expected behaviour

Modal dialogs should always appear when clicking related buttons (see below).

## Acceptance criteria

- Welcome dialog should open when clicking Getting Started button
- Confirmation dialog should open when clicking Reset Progress button
- Handbook dialog should open when clicking Handbook button
- Mission Information dialog should open when clicking MissionInfo button in level banner
- Mission Information dialog should open when selecting a level from the Welcome dialog
- Mission Information dialog should open when selecting a different level in the app header
- Levels Complete dialog should open when Level 3 is beaten
- Mission Information dialog should open when selecting "go to sandbox" from Levels Complete dialog
- Confirmation dialog should open when clicking Reset Level button
- Document Viewer dialog should open when clicking View Documents button (visible in sandbox only)

chriswilty:
## Description

There were a couple of timing-related bugs in the code, mostly due to HTMLDialogElement throwing an error if you try to open a dialog when it's already open. This was happening due to some convoluted logic in the sequence of events from Reset All Progress dialog to Welcome dialog to Mission Info dialog, including some useEffect side-effects on component update. I've simplified the logic, but also needed to remove the "click away to close dialog" behaviour, because it was causing an unrecoverable problem in which more than one click handler could become bound to the window, causing some dialogs to close as soon as they'd opened.

Resolves #862 

## Notes

- "Click away to close" behaviour has been removed
- "Escape to close" behaviour unchanged
- Tested as thoroughly as I can!

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -46,33 +46,8 @@ function App() {
 	}, [currentLevel]);
 
 	useEffect(() => {
-		if (isNewUser) openWelcomeOverlay();
+		isNewUser && openWelcomeOverlay();
 	}, [isNewUser]);
-
-	useEffect(() => {
-		// must re-bind event listener after changing overlay type
-		setTimeout(() => {
-			// Need timeout, else dialog consumes same click that
-			// opened it and closes immediately!
-			window.addEventListener('click', handleOverlayClick);
-		});
-		return () => {
-			window.removeEventListener('click', handleOverlayClick);
-		};
-	}, [overlayComponent]);
-
-	const handleOverlayClick = useCallback(
-		(event: MouseEvent) => {
-			if (
-				overlayComponent !== null &&
-				contentRef.current &&
-				!event.composedPath().includes(contentRef.current)
-			) {
-				closeOverlay();
-			}
-		},
-		[closeOverlay, contentRef, overlayComponent]
-	);
 
 	const handleEscape = useCallback(
 		(event: KeyboardEvent) => {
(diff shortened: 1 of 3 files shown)


pmarsh-scottlogic (approved):
looks good and behaves well

chriswilty:
Rebased and verified still working ok.

PR #868, opened 2024-03-20
712 Show error when unable to get document to view

pmarsh-scottlogic:
## Description

A `DocumentMeta` is a set of information about a retrieval document in the backend (e.g, uri, filename, type etc). We retrieve a list of `documentMeta`s when the `DocumentViewBox` component mounts. If we fail to retrieve these from the backend, then we show an error message.

## Screenshots

image

## Concerns

- This won't help us if the backend stops responding in between retrieving the `documentMeta`s and attempting to retrieve an individual document, but there's not a lot we can do about that given the library component.

## Checklist

Have you done the following?

- [x] Added tests

### DIFF
--- a/frontend/src/components/DocumentViewer/DocumentViewBox.test.tsx
+++ b/frontend/src/components/DocumentViewer/DocumentViewBox.test.tsx
@@ -261,4 +261,21 @@ describe('DocumentViewBox component tests', () => {
 			expect(nextButton).toBeEnabled();
 		});
 	});
+
+	describe('With zero documents', () => {
+		const documents = [] as MockDocument[];
+
+		beforeEach(() => {
+			setupMocks(documents);
+		});
+
+		test('GIVEN are zero documents THEN an error message is shown', async () => {
+			const ExpectedErrorText =
+				'Unable to fetch documents. Try again in a few minutes.';
+			renderDocumentViewBox();
+			const messageElement = await screen.findByText(ExpectedErrorText);
+
+			expect(messageElement).toBeInTheDocument();
+		});
+	});
 });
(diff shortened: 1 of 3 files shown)


frontend/src/components/DocumentViewer/DocumentViewBox.tsx
```diff
@@ -39,37 +39,44 @@ function DocumentViewBox({ closeOverlay }: { closeOverlay: () => void }) {
...
+								config={{
+									header: {
+										disableHeader: true,
+									},
+								}}
+							/>
+						</div>
+					</>
+				) : (
+					<p>Unable to fetch documents. Try again in a few minutes.</p>
```
chriswilty, line 78: `a few minutes`

It might fix itself on next refresh if it was a network blip. Maybe we can suggest closing the popup and trying again, and if the problem persists, contact your support rep? Or something like that.

frontend/src/components/DocumentViewer/DocumentViewBox.test.tsx
```diff
@@ -261,4 +261,21 @@ describe('DocumentViewBox component tests', () => {
...
+			setupMocks(documents);
+		});
+
+		test('GIVEN are zero documents THEN an error message is shown', async () => {
+			const ExpectedErrorText =
+				'Unable to fetch documents. Try again in a few minutes.';
+			renderDocumentViewBox();
+			const messageElement = await screen.findByText(ExpectedErrorText);
+
+			expect(messageElement).toBeInTheDocument();
```
chriswilty, line 278: Do we care that the document navigator is not visible, or just that the error message is visible? Probably the latter, so I think this is fine.

I think the test name is missing a "there" after GIVEN.

chriswilty (approved):
This looks fine to me.

Are we able to do anything about failure to retrieve an individual document? Or do we already handle that?

PR #872, opened 2024-03-21
645 streamline network calls for resetting level progress

pmarsh-scottlogic:
## Description

Here's what happens when you used to reset level progress:

- `openai/clear` clears the chat history for given level.
- `email/clear` clears the emails for given level.
- `openai/addHistory` adds the "Level progress reset" info message to given level's chat history

Now we just have
- `resetlevel` which clears the chat history and emails of current level and returns the resulting into `chatMessage`

## Screenshots

image

## Notes
- adds the `resetlevel` endpoint, and related service methods and models 
- adds tests for the above
- removes the `email/clear` endpoint (and therefore the whole email controller and email service and related models)
- removes the `openai/clear` endpoint (and related service and model code)
- removes tests for the above
- fixes a typo on `StartResponse`
- renames the frontend's _models/combined.ts_ to _models/apiResponse_ which @chriswilty  suggested a few weeks back

## Concerns

- not sure about the endpoint name
- not sure whether we should wrap the resulting message in an object `{resultingChatInfoMessage: //}` . I think it's better to make it clear what this message is by giving it a name in the response.

## Checklist

Have you done the following?

- [x] Added tests


backend/src/models/api/LevelResetRequest.ts
```diff
@@ -0,0 +1,15 @@
+import { Request } from 'express';
+
+import { ChatMessage } from '@src/models/chatMessage';
+import { LEVEL_NAMES } from '@src/models/level';
+
+export type LevelResetRequest = Request<
+	never,
+	{
+		resultingChatInfoMessage?: ChatMessage;
```
chriswilty, line 9: Yes this is fine as an object. Maybe as this is the response, we can name it simply "chatInfoMessage"?

backend/src/sessionRoutes.ts
```diff
@@ -92,20 +93,17 @@ router.post('/defence/deactivate', handleDefenceDeactivation);
...
 router.post('/openai/addInfoToHistory', handleAddInfoToChatHistory);
-router.post('/openai/clear', handleClearChatHistory);
 
 // model configurations
 router.post('/openai/model', handleSetModel);
 router.post('/openai/model/configure', handleConfigureModel);
 
 // reset progress for all levels
 router.post('/reset', handleResetProgress);
+router.post('/resetlevel', handleResetLevel);
```
chriswilty, line 106: I was thinking of something like this for the endpoints. Originally I considered a path param for reset level, but it didn't seem quite right, plus this is nice as it's consistent:

```
POST /restart?currentLevel=N

POST /reset?level=N
```

I also note that the /level endpoint seems a touch redundant, and could definitely use a path param, but that's a separate matter.

```
GET /level?level=1
vs
GET /level/1
```

chriswilty (commented):
This looks really good, lovely to have so much code removed! A couple of minor suggestions only.

backend/src/sessionRoutes.ts
pmarsh-scottlogic: I dislike using restart and reset to distinguish between resetting the whole game and resetting a single level - think it should be spelled out which does which. I'm also keen to keep the language consistent with the frontend. 

happy with other alternatives that do this. e.g:

```
POST /resetall?currentLevel=N

POST /reset?level=N
```

```
POST /reset/all?currentlLevel=N

POST /reset?level=n
```

chriswilty: Well, there's little difference between resetall and reset and restart, so it's going to be confusing whichever of these we would pick...

How about "resetgame" or similar?

Alternatively, we could use my initial thought about using a path param combined with yours for reset all:

```
POST /reset/all?currentLevel=N
POST /reset/{level}
```

Then it comes down to making it work in Express; likely we need them defined in the above order, so that express looks for "all" before trying a path param for level.

pmarsh-scottlogic: yeah I like this. Done

frontend/src/models/apiResponse.ts
```diff
@@ -19,8 +19,8 @@ type LoadLevelResponse = {
 	chatModel?: ChatModel;
 };
 
-type ResetLevelResponse = {
-	resultingChatInfoMessage: ChatMessageDTO;
+type chatInfoMessageResponse = {
```
chriswilty, line 22: Oops, capitalise please!

frontend/src/service/resetService.ts
```diff
@@ -28,16 +31,16 @@ async function resetAllProgress(level: number) {
 }
 
 async function resetLevelProgress(level: number) {
-	const response = await sendRequest(`resetlevel?level=${level}`, {
+	const response = await sendRequest(`${PATH}/${level}`, {
```
chriswilty, line 34: This one looks ok, but now resetAllProgress is missing the `/all` path part.

chriswilty (changes requested):
Nearly there...

chriswilty (approved):
Lovely 👍

Issue #780, opened 2024-01-18
Refactor: Shift the logic for checking win condition


pmarsh-scottlogic:
Question: do we want this? I think it will make things easier to follow.

this goes along with #761. Do it before or after, but not at the same time.

## Description

**summary**: _move the win condition logic out from deep within the call stack, to encourage a more flat structure, and be better for single responsibility._

Right now, a user sends a message, and to see if that message has caused the level to be completed, we have to look right at the bottom of the call stack: `handleChatToGPT(...)` => `handleLow[orHigher]LevelChat(...)` => `chatGptSendMessage(...)` => `getFinalReplyAfterAllToolCalls(...)` => `performToolCalls(...)` => `chatGptCallFunction(...)` => `sendEmail(...)` => `checkLevelWinCondition(...)`.

I think we should separate the logic for checking the win condition from the logic that deals with processing the user's message and getting a reply. Something like
```typescript
function handleChatToGPT(...) {
  const reply = handleLow[orHigher]LevelChat(...); // reply object includes a list of sent emails
  chatReponse.wonLevel = checkLevelWinCondition(reply.sentEmails);
  res.send(chatResponse);
```
Which will leave `chatGptSendMessage(...)` responsible for one fewer thing. Solid.

## Acceptance Criteria
refactor, so no behaviour change. Should be covered by dev tests...

pmarsh-scottlogic:
this might be able to skip testing? Or maybe just regressions on winning a level. Test also winning a level when the user's message is blocked

PR #873, opened 2024-03-21
828 steamline win level info message

pmarsh-scottlogic:
## Description

Before, when the user would send a message that would cause the level to be won, we would check if `wonLevel` in the backend's response was `true`, and if so, we would add a congratulations message to the chat History, causing an extra API call.

Now, we return the congratulations messagge (`wonLevelMessage`) in the chat response. The front end simply appends it to the history if it is there, and can use this to determine if the user has won the level.

## Screenshot

image

## Notes
also does these things:
- renames the `chatMessageType` `LEVEL_INFO` to `LEVEL_COMPLETE`
- removes the information header from the congratulations message. Adds party poppers instead

## Concerns

- Now in the backend, the type `ChatHttpResponse` contains the `wonLevel` boolean property. It gets marked as true at the points of the send email function call. At the end of `handleChatToGPT` in the _chatController_, we check `wonLevel`. If we have won the level, then we add the congratulations message to the history and return it to the frontend. The frontend only needs the congratualtions message to determine that a level has been won, making wonLevel redundant in the frontend. This is untidy, but not necessarily problematic. This will be made irrelevant by #780 

## Checklist

Have you done the following?

- [x] Added tests


frontend/src/components/ChatBox/ChatBox.tsx
```diff
@@ -152,19 +141,13 @@ function ChatBox({
...
-				type: 'LEVEL_INFO',
-				message: successMessage,
-			});
-			// asynchronously add the message to the chat history
-			void chatService.addInfoMessageToChatHistory(
-				successMessage,
-				'LEVEL_INFO',
-				currentLevel
+			const levelCompleteMessage = makeChatMessageFromDTO(
+				response.wonLevelMessage
```
chriswilty, line 147: This is maybe something that will emerge after all your changes, but my feeling is that it's fine to return the "wonLevel" flag as your indicator, and the actual "won level" message could arrive as an info message.

Reason being, we already have an optional info message in the response that is added to the chat panel ("transformedMessage" though it's currently a string, not an actual ChatMessage), and so adding another optional field for another info message to the response seems like we are making our API a little too specific. Maybe we want to return a list of (zero or more) info messages to add to the chat?

Likely not something to tackle here, but something to consider as you continue with this work.

frontend/src/components/ChatBox/ChatBoxMessage/MessageBubble.tsx
```diff
@@ -48,12 +48,6 @@ function MessageBubble({
 	return (
 		// eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
 		<section className={className} lang="en" tabIndex={0}>
-			{message.type === 'LEVEL_INFO' && (
-				<>
-					<p className="header">Information</p>
-					<span className="visually-hidden"> message: </span>
-				</>
```
chriswilty, line 55: Yeah that seems redundant, happy for it to go.

chriswilty (changes requested):
Minor suggestions only, and one to chew on.

frontend/src/components/ChatBox/ChatBox.tsx
pmarsh-scottlogic: I like the suggestion. Like, what happens if the message gets transformed, but that message would also win the level? My current solution doesn't support that

pmarsh-scottlogic: nevermind, yes it does. But your idea is sleeker

pmarsh-scottlogic: that can be a different PR :)

chriswilty (approved):
Lovely!

frontend/src/components/ChatBox/ChatBox.tsx
pmarsh-scottlogic: https://github.com/orgs/ScottLogic/projects/37/views/2?pane=issue&itemId=57715194 made a ticket for it

pmarsh-scottlogic:
I need to solve a problem where we are not sending the congratulations message if there is an OpenAI error after the winning email has been sent

pmarsh-scottlogic:
I've decided the gain is not worth the cost. In the startlingly unlikely event that the user would win a level with a sent email, but open has an error after the function call, then the user will not win the level.

PR #874, opened 2024-03-22
828 streamline configure defence info message network call

pmarsh-scottlogic:
⚠ Note, don't believe the branch name!

## Description

Before, when the user would configure a defence, we would send a request to the backend. If the backend successfully configured the defence, then the frontend would send a further request to add the info message to the chat history, eg 'character limit defence updated'.

This PR changes it so that when the backend successfully configures a defence, the backend generates the info message and appends it to the history, then forwards it on to the frontend.

## Concerns

- similar to #872 I'm not certain we want to be wrapping the resultingChatInfoMessage in the response object just so that we can name it.

## Checklist

Have you done the following?

- [x] Added tests

### DIFF
--- a/backend/src/models/api/DefenceConfigureRequest.ts
+++ b/backend/src/models/api/DefenceConfigureRequest.ts
@@ -8,7 +8,7 @@ export type DefenceConfigureRequest = Request<
 	never,
 	| string
 	| {
-			resultingChatInfoMessage: ChatInfoMessage;
+			chatInfoMessage: ChatInfoMessage;
 	  },
 	{
 		config?: DefenceConfigItem[];
--- a/backend/src/models/api/DefenceConfigureRequest.ts
+++ b/backend/src/models/api/DefenceConfigureRequest.ts
@@ -1,11 +1,15 @@
 import { Request } from 'express';
 
+import { ChatInfoMessage } from '@src/models/chatMessage';
 import { DEFENCE_ID, DefenceConfigItem } from '@src/models/defence';
 import { LEVEL_NAMES } from '@src/models/level';
 
 export type DefenceConfigureRequest = Request<
 	never,
-	null | string,
+	| string
+	| {
+			resultingChatInfoMessage: ChatInfoMessage;
+	  },
 	{
 		config?: DefenceConfigItem[];
 		defenceId?: DEFENCE_ID;
--- a/frontend/src/service/defenceService.ts
+++ b/frontend/src/service/defenceService.ts
@@ -66,12 +66,12 @@ async function configureDefence(
 		body: JSON.stringify({ defenceId, config, level }),
 	});
 
-	const { resultingChatInfoMessage } =
+	if (response.status !== 200) return null;
+
+	const { chatInfoMessage } =
 		(await response.json()) as ConfigureDefenceResponse;
 
-	return response.status === 200
-		? makeChatMessageFromDTO(resultingChatInfoMessage)
-		: null;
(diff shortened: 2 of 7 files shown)


backend/src/models/api/DefenceConfigureRequest.ts
```diff
@@ -1,11 +1,15 @@
...
 import { DEFENCE_ID, DefenceConfigItem } from '@src/models/defence';
 import { LEVEL_NAMES } from '@src/models/level';
 
 export type DefenceConfigureRequest = Request<
 	never,
-	null | string,
+	| string
+	| {
+			resultingChatInfoMessage: ChatInfoMessage;
+	  },
```
chriswilty, line 12: Does prettier not have something to say about these whitespace indents?

backend/src/models/api/DefenceConfigureRequest.ts
```diff
@@ -1,11 +1,15 @@
...
+import { ChatInfoMessage } from '@src/models/chatMessage';
 import { DEFENCE_ID, DefenceConfigItem } from '@src/models/defence';
 import { LEVEL_NAMES } from '@src/models/level';
 
 export type DefenceConfigureRequest = Request<
 	never,
-	null | string,
+	| string
+	| {
+			resultingChatInfoMessage: ChatInfoMessage;
```
chriswilty, line 11: `string` or `object` - is that cos we have some error responses with a string body?

What do we do with those error responses? I'm now checking your changes in defenceService...

frontend/src/service/defenceService.ts
```diff
@@ -54,15 +57,21 @@ async function configureDefence(
...
 		method: 'POST',
 		headers: {
 			'Content-Type': 'application/json',
 		},
 		body: JSON.stringify({ defenceId, config, level }),
 	});
-	return response.status === 200;
+
+	const { resultingChatInfoMessage } =
+		(await response.json()) as ConfigureDefenceResponse;
```
chriswilty, line 70: This could be problematic because the response type in the case of errors is not application/json, it's text/plain. So I expect `response.json()` to fail in that case. Did you try it?

Better to check the status first, and if Ok then parse the JSON and make your DTO. And if it's not Ok, what happens to the error message? Looks like we are swallowing that currently.

chriswilty (changes requested):
Looks like your changes have uncovered some unhandled error cases - or at least, swallowed ones. Could be a nice win for our final week!

backend/src/models/api/DefenceConfigureRequest.ts
pmarsh-scottlogic: Yes, sometimes we return an error message a simple string.  Also yes, we swallow those. I suppose they remain helpful because you can see them in the browser inspector.

A note on these request types: This second part (ResBody) seems entirely redundant. TypeScript doesn't seem to enforce the types at all when you call `res.send(...)`. Are we missing something?

frontend/src/service/defenceService.ts
pmarsh-scottlogic: Yup, good call to check the status first. And yes, the message gets swallowed.

pmarsh-scottlogic: not-swallowing the errors is a wider UI problem I reckon

chriswilty: It's that "unhappy paths" issue come back to bite us 😁

chriswilty (approved):
Beautiful!

PR #875, opened 2024-03-22
828 streamline chat model configuration info message network call

pmarsh-scottlogic:
## Description

Before, when the user would configure the chatModel, we would send a request to the backend. If the backend successfully configured the model, then the frontend would send a further request to add the info message to the chat history, eg 'changed frequencyPenalty to 0.6'.

This PR changes it so that when the backend successfully configures the chatModel, the backend generates the info message and appends it to the history, then forwards it on to the frontend.

## Notes
- `handleConfigureModel` now generates an info message, appends it to the chat history and forwards it to the frontend.
- renames `MODEL_CONFIG` enum to `MODEL_CONFIG_ID` and converts it to a string literal type, on frontend and backend.
- adds detailed validation to `handleConfigureModel`.
- adds unit tests for `handleConfigureModel`.

## Checklist

Have you done the following?

- [x] Added tests


frontend/src/models/chat.ts
```diff
@@ -15,27 +15,30 @@ type CHAT_MESSAGE_TYPE =
...
-interface ChatModelConfigurations {
+type ChatModelConfigurations = {
 	temperature: number;
 	topP: number;
 	frequencyPenalty: number;
 	presencePenalty: number;
-}
+};
+
+// these must match the above type because they are used for indexing
```
chriswilty, line 30: Same suggestion as for backend code.

frontend/src/service/chatService.ts
```diff
@@ -75,15 +76,19 @@ async function setGptModel(model: string): Promise<boolean> {
...
-): Promise<boolean> {
+): Promise<ChatMessage | null> {
 	const response = await sendRequest(`${PATH}model/configure`, {
 		method: 'POST',
 		headers: { 'Content-Type': 'application/json' },
 		body: JSON.stringify({ configId, value }),
 	});
-	return response.status === 200;
+	const { resultingChatInfoMessage } =
+		(await response.json()) as ConfigureGptModelResponse;
```
chriswilty, line 88: Yes this is the same problem as for defences: the response might not be JSON, plus we are swallowing update errors.

chriswilty (commented):
Much the same as for defence config updates, nice work but I _think_ I see a problem.

backend/src/models/chat.ts
```diff
@@ -17,24 +17,27 @@ enum CHAT_MODELS {
...
+type ChatModelConfiguration = {
 	temperature: number;
 	topP: number;
 	frequencyPenalty: number;
 	presencePenalty: number;
-}
+};
+
+// these must match the above type because they are used for indexing
+const modelConfigIds = [
```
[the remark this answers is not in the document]
pmarsh-scottlogic: Aha! Just what I was looking for :)

frontend/src/service/chatService.ts
pmarsh-scottlogic: cool, let's resolve that there and assume that it'll apply here as well

backend/src/models/chat.ts
```diff
@@ -39,6 +31,10 @@ const modelConfigIds = [
 
 type MODEL_CONFIG_ID = (typeof modelConfigIds)[number];
 
+type ChatModelConfigurations = {
```
chriswilty, line 34: There must be a reason you pluralised this, right? Is there a clash with another type, or did you feel it made more sense as a plural?

frontend/src/components/ModelBox/ModelConfigurationSlider.tsx
```diff
@@ -1,6 +1,7 @@
 import { Slider } from '@mui/material';
 import { useEffect, useState } from 'react';
 
+import DetailElement from '@src/components/ThemedButtons/DetailElement';
```
chriswilty, line 4: Hmm... Was this a merge thing gone askew, and corrected here?

frontend/src/service/chatService.ts
```diff
@@ -84,11 +84,13 @@ async function configureGptModel(
...
 		body: JSON.stringify({ configId, value }),
 	});
-	const { resultingChatInfoMessage } =
-		(await response.json()) as ConfigureGptModelResponse;
-	return response.status === 200
-		? makeChatMessageFromDTO(resultingChatInfoMessage)
-		: null;
+
+	if (response.status !== 200) return null;
+	else {
```
chriswilty, line 89: Strictly speaking you don't need the else.

chriswilty (approved):
Nice work 👌 

A couple of random thoughts but nothing that should prevent merging.

backend/src/models/chat.ts
pmarsh-scottlogic: I prefer it that way.

If each of top p, `frequency penalty etc are one configuration, then the set of all of them should be called configurations.

chriswilty: That's fine, but then export it as that - currently you're exporting it as ChatModelConfiguration.

chriswilty (changes requested):
One teeny thing, else good to go 👍

chriswilty (approved):
Get it in! ⚡

PR #876, opened 2024-03-25
780 refactor shift the logic for checking win condition

pmarsh-scottlogic:
## Description

Moves the code that checks whether the user has won. It used to live within the `sendEmail`, deep within the call stack for sending a chat message. I've brought it right out to the top level.

## Notes

- moves `checkLevelWinCondition` and `checkSubjectAndBodyContains` from _email.ts_ to a new file _winCondition.ts_.
- calls `checkLevelWinCondition` from `handleChatToGPT`.
- `wonLevel` no longer being passed all the way back up through the call stack.
- converts a few interface definitions to type definitions.
- testing for winning a level is now handled in the unit tests for _winCondition.ts_ and _chatController.ts_.

## Checklist

Have you done the following?

- [ ] Linked the relevant Issue
- [ ] Added tests

### DIFF
--- a/backend/src/openai.ts
+++ b/backend/src/openai.ts
@@ -160,10 +160,7 @@ async function handleAskQuestionFunction(
 	}
 }
 
-function handleSendEmailFunction(
-	functionCallArgs: string | undefined,
-	currentLevel: LEVEL_NAMES
-) {
+function handleSendEmailFunction(functionCallArgs: string | undefined) {
 	if (functionCallArgs) {
 		const params = JSON.parse(functionCallArgs) as FunctionSendEmailParams;
 		console.debug('Send email params: ', JSON.stringify(params));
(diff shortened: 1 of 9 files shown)


backend/src/openai.ts
```diff
@@ -423,7 +409,6 @@ async function chatGptSendMessage(
 
 	const chatResponse: ChatResponse = {
 		completion: finalToolCallResponse.gptReply.completion,
-		wonLevel: finalToolCallResponse.wonLevel,
```
chriswilty, line 426: I love how many places this has been removed from 🥳

backend/src/winCondition.ts
```diff
@@ -0,0 +1,64 @@
...
+					'20,000,000',
+					'20.000.000',
+				])
+			);
+		default:
+			return false;
+	}
+}
+
+function checkLevelWinCondition(
```
chriswilty, line 55: Would "isLevelWon" be a nicer name? "check" doesn't as easily convey what we expect, which is a boolean result.

backend/test/unit/controller/chatController.test.ts
```diff
@@ -737,6 +748,131 @@ describe('handleChatToGPT unit tests', () => {
...
+
+			mockCheckLevelWinCondition.mockReturnValueOnce(true);
+
+			await handleChatToGPT(req, res);
+
+			expect(res.send).toHaveBeenCalledWith(
+				expect.objectContaining({ wonLevel: true })
+			);
+		});
+	});
```
chriswilty, line 875: Nice tests, really clear 👍

chriswilty (approved):
Lovely work, so much cleaner 💎 

As I seem unable to let a PR through without a single comment 😅  I've made one small suggestion, but feel free to ignore it!

backend/src/winCondition.ts
pmarsh-scottlogic: I was thinking this too actually

PR #877, opened 2024-03-25
Docker and backend changes to support prod-lite mode

### LINKED ISSUE #867: Simplified route to internal enterprise deployment
# Feature Request

## Description

There should be minimal overhead for an enterprise wishing to deploy this to their internal users. Currently there's only a local deployment model and a cloud-hosted version optimised for SL's use case.

## Additional context

Ideally this could be made as simple as build this *single* container from source/download this image (depending on security preferences of enterprise), and run it with these instructions (expose port, provide API secret, etc.).

## Acceptance criteria

**GIVEN** an enterprise is keen to deploy this for their internal users
**WHEN** they look for installation instructions
**THEN** they find instructions which are as simple as possible with minimal assumptions about their infrastructure

chriswilty:
## Description

Enables running both API and UI through the Express server, in "production lite" mode.

From project root:

```
// Build UI, if not already built or if not up to date with latest changes:
npm run build

// Run UI and API together:
npm start
```

Resolves #867 

## Screenshots

Hosted on localhost:3000, and API reached on localhost:3000/api

image

## Notes

- Express server hosts UI as well as API, for running a built / bundled app
- Running in development mode is unchanged: spin up API and UI separately
- An additional Dockerfile is also provided, for running UI and API in same container
- Original Dockerfile in backend directory remains, for running just the API in a container

## Concerns

- VITE_BACKEND_URL in `frontend/.env.example` has changed, so you will need to update your `frontend/.env` to match!

## Checklist

Have you done the following?

- [ ] Added tests


.gitignore
```diff
@@ -77,11 +77,8 @@ web_modules/
 .yarn-integrity
 
 # dotenv environment variable files
-.env
-.env.development.local
-.env.test.local
-.env.production.local
-.env.local
+.env*
+!backend/.env.example
```
pmarsh-scottlogic, line 81: how come we don't need to explicitly include `frontend/.env.example`?

pmarsh-scottlogic (changes requested):
All seems sensible enough (not that I can claim to fully understand it)! Ran things locally with no problem. Didn't test docker because I don't want to faff about installing. Approved, but there are a couple of comments.

.gitignore
chriswilty: Whoops!

chriswilty:
@pmarsh-scottlogic Ok, this one should be good for a final sweep.

.gitignore
```diff
@@ -22,139 +16,43 @@ pids
...
-
-node_modules
-
-*.sublime*
-
-psd
-thumb
-sketch
-
-# End of https://www.toptal.com/developers/gitignore/api/react,node
```
pmarsh-scottlogic, line 160: this file wants a new line at the end  of it

backend/src/server/uiRoutes.ts
```diff
@@ -0,0 +1,14 @@
+import express from 'express';
+import { fileURLToPath } from 'node:url';
+
+import { importMetaUrl } from './importMetaUtils';
+
+export default express.Router().use(
+	express.static(
+		fileURLToPath(new URL('../../../frontend/dist', importMetaUrl())),
```
pmarsh-scottlogic, line 8:
```suggestion
		fileURLToPath(new URL('@src/frontend/dist', importMetaUrl())),
```
does this work?

pmarsh-scottlogic (approved):
looks good. Got 2 comments, both of which are not blocking me from approving

backend/src/server/uiRoutes.ts
chriswilty: Nope, because `@src` would be relative to _backend_, and we are needing the frontend build.

.gitignore
chriswilty: How strange, it has a newline in my local copy and it doesn't come up as a change.

PR #881, opened 2024-03-26
Mode Selection Buttons updated

dhinrichs-scottlogic:
# Description
When you select between two levels (Level 1 vs Sandbox or Level 3 vs Sandbox), the button for the level that you are currently on is differently styled to the one that you are not on. This is not very clear, so now we want the two buttons to just look the same.

Also added on a tiny `<div>` to `<nav>` change onto this PR as discussed

## Acceptance criteria
GIVEN a user is on a modal that lets you switch between two levels (Getting started & Congratulations)
WHEN they look at the two level button options
THEN they look the same

**Additional**: Level 1 2 3 Sandbox still behave as they did before but are now in a `nav` container element.  

## Screenshot before
Image

## Screenshot expected outcome
Image

## Checklist

Have you done the following?

- [ ] Added tests

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -72,7 +72,6 @@ function App() {
 		setIsNewUser(false);
 		openOverlay(
 			<OverlayWelcome
-				currentLevel={currentLevel}
 				setStartLevel={(level: LEVEL_NAMES) => {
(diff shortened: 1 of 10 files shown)


pmarsh-scottlogic (approved):
beautiful

PR #882, opened 2024-03-27
Level 1 mission info conversation & modal zoom

dhinrichs-scottlogic:
## Description
- Updated the Level1 mission info conversation text. 
- Changed the responsiveness of the multipage overlays so that the header is less squished at 150% zoom.
- Fixed alignment of the two buttons in short mission info and reduced padding for better zoom

## Screenshots
image

The two level 3 buttons become columnar (it's a technical term ;) ) at 200% zoom
image

## Checklist

Have you done the following?

- [ ] Added tests


chriswilty:
Hmm. Do the "columnar" buttons save us any vertical space?

dhinrichs-scottlogic:
> Hmm. Do the "columnar" buttons save us any vertical space?

They do indeed. 
image

chriswilty:
> > Hmm. Do the "columnar" buttons save us any vertical space?
> > @chriswilty
> > They do indeed.
> > image

Here's another suggestion, don't immediately dismiss it 😅
Do we really need to see the full mission info after completing level 3? How about if we simply replaced the full mission info with the Congrats button? Why would a user need to see the full level 3 mission info after beating the level? If they want to try the level again they can reset it, or if they move to a different level then back again, they'll see the dialog. I think we're making life difficult for ourselves, and those stacked buttons look pretty ugly...

frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
```diff
@@ -2,7 +2,7 @@
 	display: flex;
 	gap: 1.5rem;
 	align-items: center;
-	padding: 0.5rem 1.5rem;
+	padding: 0 1.5rem;
```
chriswilty, line 5: I think we leave this at 0.5rem top and bottom. On the smallest breakpoint I also knocked down the gap on this class from 1.5 to 1.25rem. Could possibly go down to 1rem, unsure.

frontend/src/components/LevelMissionInfoBanner/LevelMissionInfoBanner.css
```diff
@@ -38,11 +43,22 @@
...
 	color: var(--action-button-background-colour);
 	font: 1rem serif;
 	font-weight: bold;
 	aspect-ratio: 1;
 }
+
+@media (width < 62.5rem) {
+	.level-mission-info-banner .banner-button-container {
+		flex-direction: column;
+		gap: 0.125rem;
```
chriswilty, line 58: I think this can stay at 0.5rem, seems little to gain from squishing the buttons together.

chriswilty (changes requested):
Looking good, yes I think on reflection we're ok to use the stacked buttons, but I suggest adjusting the gaps a little.

chriswilty (approved):
Lovely 👌

PR #883, opened 2024-03-28
828 streamline change chat model request

pmarsh-scottlogic:
## Description

Previously we would set the chat model in the backend. If that was sucessful, then the frontend would send a further request to add an info message to the history. Now the backend generates this info message and forwards it to the frontend.

## Notes

- Changes `CHAT_MODEL` enum to `CHAT_MODEL_ID` string literal type.
- Adds this type to the frontend for type safety
- Adds extra validation to `handleSetModel` backend method, to check that the model is valid.
- Adds unit tests for `handleSetModel`

## Checklist

Have you done the following?

- [x] Added tests


backend/src/controller/modelController.ts
```diff
@@ -13,14 +17,32 @@ function handleSetModel(req: OpenAiSetModelRequest, res: Response) {
...
+		req.body.configuration ?? req.session.chatModel.configuration;
+	req.session.chatModel = { id: model, configuration };
+	console.debug('GPT model set:', JSON.stringify(req.session.chatModel));
+
+	const chatInfoMessage = {
+		infoMessage: `changed model to ${model}`,
+		chatMessageType: 'GENERIC_INFO',
+	} as ChatInfoMessage;
+	// for now, the chat model only changes in the sandbox level
+	req.session.levelState[LEVEL_NAMES.SANDBOX].chatHistory =
```
chriswilty, line 39: Yep that's fair. Would it help to send the level as part of the request, or can we leave that for another day / an enhancement when needed?

backend/src/models/chat.ts
```diff
@@ -7,18 +7,20 @@ import { ChatInfoMessage, ChatMessage } from './chatMessage';
...
-enum CHAT_MODELS {
-	GPT_4_TURBO = 'gpt-4-1106-preview',
-	GPT_4 = 'gpt-4',
-	GPT_4_0613 = 'gpt-4-0613',
-	GPT_3_5_TURBO = 'gpt-3.5-turbo',
-	GPT_3_5_TURBO_0613 = 'gpt-3.5-turbo-0613',
-	GPT_3_5_TURBO_16K = 'gpt-3.5-turbo-16k',
-	GPT_3_5_TURBO_16K_0613 = 'gpt-3.5-turbo-16k-0613',
-}
+const chatModelIds = [
```
chriswilty, line 10: ❤️

backend/src/utils/token.ts
```diff
@@ -4,18 +4,18 @@ import {
...
-	[CHAT_MODELS.GPT_3_5_TURBO_16K]: 16385,
-	[CHAT_MODELS.GPT_3_5_TURBO_16K_0613]: 16385,
+	['gpt4-turbo']: 128000,
+	['gpt-4']: 8192,
+	['gpt-4-1106-preview']: 128000,
+	['gpt-4-0613']: 8192,
+	['gpt-3.5-turbo']: 4097,
+	['gpt-3.5-turbo-0613']: 4097,
+	['gpt-3.5-turbo-16k']: 16385,
+	['gpt-3.5-turbo-16k-0613']: 16385,
```
chriswilty, line 18: Out of interest, are these still valid? I reckon not... Look at gpt-3.5-turbo for example.

chriswilty (changes requested):
I think our tokens-per-model have not been kept up to date, but otherwise this all looks fine 👌

backend/src/controller/modelController.ts
pmarsh-scottlogic: it would be nicer and more consistent, but would make no real difference. I wasn't gonna bother

backend/src/utils/token.ts
pmarsh-scottlogic: that was the only invalid one. Though 3 of them will be deprecated.

chriswilty (approved):
Niiiice 👌

PR #884, opened 2024-03-28
828 streamline network call for toggle defence

pmarsh-scottlogic:
## Description

Previously we would activate/deactivate  in the backend. If that was sucessful, then the frontend would send a further request to add an info message to the history. Now the backend generates this info message and forwards it to the frontend.

## Concerns

- There's no plan for if the request fails. Not necessarily in scope of ticket? i.e. don't want to deal with it right now

## Checklist

Have you done the following?

- [x] Added tests


frontend/src/service/defenceService.ts
```diff
@@ -50,7 +50,12 @@ async function toggleDefence(
 		},
 		body: JSON.stringify({ defenceId, level }),
 	});
-	return response.status === 200;
+	if (response.status !== 200) return null;
+
+	const { chatInfoMessage } =
+		(await response.json()) as ChatInfoMessageResponse;
+
+	return makeChatMessageFromDTO(chatInfoMessage);
```
chriswilty, line 58: The lovely thing about returning a ChatMessage from all these service functions is that it'd be really easy to return an "error" chat message to go into the chat panel, instead of `null` to be ignored. If only we have time....

chriswilty (approved):
Sweet 🍯

PR #892, opened 2024-04-09
Update main branch for Release 1.0.0

### LINKED ISSUE #890: Release 1.0
Time to release version 1.0, now the main team are moving onto other things.

Note that we may want to release a patch version shortly after, with all vulnerabilities tackled. For example, langchain lib is on an old 0.0.* version and we should be using latest 0.1.* now, but that could involve breaking changes, so we'll do it after the major release.

chriswilty:
## Description

Time to cut release 1.0.0 - this means bringing main up to date with dev. On the source branch I have cherry-picked all the commits since last merge; this is unfortunately necessary as the previous merges to main were accidentally squashed.

In theory, we could correct this for future releases by
1. removing protections from main branch,
2. deleting main branch
3. recreating main branch at current dev
4. reapplying branch protections

If we don't do it now, before creating the tag, we'll never do it 😓 

Resolves #890 

## Notes

- Checked linting/formatting, front and back
- Ran tests front and back
- Ran application locally and manually tested in full(ish)

## Checklist

Have you done the following?

- [ ] Added tests


chriswilty:
I am going to abandon this in favour of reworking _main_ branch so that it exactly mirrors commits on dev. This would not normally be done, however we don't have any tags on main nor any other branches based on main, so it's safe.

I will need to temporarily remove branch protections on main, remove the two squashed commits, then open a PR to merge dev back into main without squashing. It seems that in order to preserve committer history, we need to use GitHub's default "merge without fast-forward" strategy which will add a merge commit, but this is ok as it will be empty, thus not cause any merge conflicts on future merges back into main.

PR #903, opened 2024-04-23
Use all vectorstore documents in QA LLM context

### LINKED ISSUE #902: Investigate why Q&A bot will not find all hits in documents
Our Q&A bot searches the documents of the current level for answers, when the user asks a question. However, it is proving to be almost impossible to retrieve all the information a user wants in a single chat.

For example:
- I ask the bot to tell me the salaries of all employees, knowing there is a document (management_info.csv) listing nine manager employees and their salaries.
- Bot returns me a seemingly random selection of just three or four employees from the list.
- I ask for others, and the bot insists that's all it knows about
- If I keep insisting there are more, then sometimes an extra one or two will be found, but sometimes even fewer are retrieved, and each time the bot insists there are no more.
- Sometimes it returns me just one, and insists there are no more.

We need to investigate what's going on here, because if we cannot retrieve all the information we are looking for without any defences enabled, then Sandbox level will prove to be a frustrating experience.

Places to start:
- Does the Q&A bot actually return all the info we seek, but the chatbot then ignores that when generating its final response?
- Are there any limits on searches that are configurable, and that are currently preventing all hits being returned, or all documents from being included in a search?

Stick a load more logging in place to begin with, and see what each bot chat or tool call returns.

**Acceptance Criteria**
GIVEN I am on Sandbox level
AND no defences are enabled
WHEN I ask the bot for salaries of all known employees
THEN all nine entries from file `backend/resources/documents/common/management_info.csv` are given in the chat response

As bot responses are non-deterministic, you might need to check this a few times to see if you can get it to reveal all nine salaries. An alternative might be to ask it for the names of all employees, to see if it returns them all, and then to ask for the salaries of each of those.

chriswilty:
## Description

Found and squished a bug in our Q&A process: we were not passing all documents in chat context to the Q&A LLM, as the default documents to retrieve from is 4 and we were not overriding this. This caused an odd symptom in which the bot was seemingly unable to answer a question about employees or salaries in full - instead one needed to piece together a complete answer from several consecutive questions, and even then you'd not know you were in receipt of the complete answer unless you knew what you were looking for.

The fix is simple, though there are related test changes, plus an incorrect comment and lint-ignore statement were also tackled here.

#Resolves #902

## Screenshots

All salary info found in one call:

image

## Checklist

Have you done the following?

- [ ] ~Added tests~ Corrected existing tests

### DIFF
--- a/backend/src/document.ts
+++ b/backend/src/document.ts
@@ -1,7 +1,9 @@
 import { OpenAIEmbeddings } from '@langchain/openai';
 import { CSVLoader } from 'langchain/document_loaders/fs/csv';
-import { DirectoryLoader } from 'langchain/document_loaders/fs/directory';
-import { PDFLoader } from 'langchain/document_loaders/fs/pdf';
+import {
+	DirectoryLoader,
+	UnknownHandling,
+} from 'langchain/document_loaders/fs/directory';
 import { TextLoader } from 'langchain/document_loaders/fs/text';
 import { RecursiveCharacterTextSplitter } from 'langchain/text_splitter';
 import { MemoryVectorStore } from 'langchain/vectorstores/memory';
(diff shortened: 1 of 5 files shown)


backend/src/document.ts
```diff
@@ -14,25 +16,20 @@ import { LEVEL_NAMES } from './models/level';
 async function getDocuments(filePath: string) {
 	console.debug(`Loading documents from: ${filePath}`);
 
-	const loader: DirectoryLoader = new DirectoryLoader(filePath, {
-		'.pdf': (path: string) => new PDFLoader(path),
```
chriswilty, line 18: We don't have PDFs so let's remove this one.

backend/src/document.ts
```diff
@@ -14,25 +16,20 @@ import { LEVEL_NAMES } from './models/level';
...
 	console.debug(`${docs.length} documents found`);
 
-	// split the documents into chunks
-	const textSplitter = new RecursiveCharacterTextSplitter({
-		chunkSize: 1000,
-		chunkOverlap: 0,
-	});
-
-	// need to disable the linting rule here
-	// as it's possible for the splitDocuments method to return undefined
```
chriswilty, line 32: This is no longer true, so we can simplify the code here.

backend/src/langchain.ts
```diff
@@ -50,9 +50,11 @@ function initQAModel(level: LEVEL_NAMES, Prompt: string) {
 		'QA prompt template'
 	);
 	console.debug(`QA chain initialised with model: ${modelName}`);
-	return RetrievalQAChain.fromLLM(model, documentVectors.asRetriever(), {
-		prompt: promptTemplate,
-	});
+	return RetrievalQAChain.fromLLM(
+		model,
+		documentVectors.asRetriever(documentVectors.memoryVectors.length),
```
chriswilty, line 55: This is the actual fix! Use all docs from the vector store in our retrieval context.

PR #916, opened 2024-07-09
913: Document Viewer improvements

### LINKED ISSUE #913: Doc Viewer shows "unable to fetch documents" error during initial load
# Bug report

## Description

On opening the Document Viewer in sandbox, the available document list is fetched from the server. But while we wait for that to load, we see error message "Unable to fetch documents..." until the request completes. This is obviously misleading. Instead we should probably see a loading indicator, and an error message only if the request failed or returned with an empty document list.

Image

The problem is exacerbated because the component is unmounted on close, meaning a request is made every time the viewer is opened. We might therefore consider lifting the "available documents" state into a parent component, and passing it in as a prop. Currently the App component is the parent of the viewer, so the state would need to live there, and probably be fetched when Sandbox is first opened.

Alternatively, we could send the available documents list in the level data response. Level data are fetched from MainComponent rather than App, so the function to open the document viewer would need to be moved into there. This would be the more efficient solution.

**Update**
The "alternative" solution above will be implemented for reasons of efficiency.

## Reproduction steps

Steps to reproduce the behaviour:

1. Open the Sandbox level
2. Click on View Documents in the left side panel
3. Message is visible: "Unable to fetch documents ..."
4. After a short pause, the above message is replaced by the first document.

## Expected behaviour

Error message does not display on initial fetch; instead, a loading indicator is displayed if the docs request is in progress.

## Acceptance criteria

**GIVEN** I am on Sandbox level
**WHEN** I open the Document Viewer
**THEN** the first document loads immediately

chriswilty:
## Description

Documents metadata are now returned in the (sandbox) level data payload. This results in two improvements:
- no need to request metadata every time the Document Viewer is opened
- first document can be requested as soon as the viewer opens, so we never need to see a loading indicator for the metadata fetch

_This might be easiest viewed per commit, instead of all in one go._

resolves #913 

## Notes

- There were case-sensitivity typos in DocumentMeta model properties, causing a redundant HEAD request for every document fetched; these have been corrected, so now DocViewer code can use `fileType` to know what renderer to use, and the HEAD request is not needed
- Improvements made to frontend service code, to remove some duplicated boilerplate - now we have separate `get` and `post` functions (copied over from the cloud infra feature branch)
- Stronger type safety for responses in backend controller code

## Checklist

Have you done the following?

- [x] Added tests


backend/src/document.ts
```diff
@@ -58,19 +58,16 @@ function getSandboxDocumentMetas() {
...
-
-	fs.readdirSync(filepath).forEach((file) => {
+	return fs.readdirSync(filepath).map((file) => {
 		const fileType = file.split('.').pop() ?? '';
-		documentMetas.push({
-			filename: file,
-			filetype: fileType === 'csv' ? 'text/csv' : fileType,
+		return {
+			fileName: file,
+			fileType: fileType === 'csv' ? 'text/csv' : fileType,
```
chriswilty, line 67: Typo fixes! This was causing redundant HEAD requests in the UI.

backend/src/server/nonSessionRoutes.ts
```diff
@@ -11,8 +10,13 @@ export default express
 	.use(
 		'/documents',
 		express.static(
-			fileURLToPath(new URL('../../resources/documents', importMetaUrl()))
+			fileURLToPath(new URL('../../resources/documents', importMetaUrl())),
+			{
+				cacheControl: true,
+				maxAge: 604800000, // 7 days as millis
```
chriswilty, line 16: Documents rarely change, so we can tell the browser to cache them.

frontend/src/components/DocumentViewer/DocumentViewBox.tsx
```diff
@@ -1,36 +1,25 @@
...
-function DocumentViewBox({ closeOverlay }: { closeOverlay: () => void }) {
-	const [documentMetas, setDocumentMetas] = useState<DocumentMeta[]>(emptyList);
+function DocumentViewBox({
+	documents,
+	closeOverlay,
+}: {
+	documents?: IDocument[];
+	closeOverlay: () => void;
+}) {
 	const [documentIndex, setDocumentIndex] = useState<number>(0);
```
chriswilty, line 22: Interesting story: I tried moving this "current document index" into the parent component, so that the viewer would remember where you were when reopening. Didn't work.

Why? Because even though the JSX that declares our DocumentViewBox belongs to the parent (MainComponent.tsx), we pass that to Overlay component which is rendered by the grandparent (App.tsx). We can store the state in the parent, and when calling setDocumentIndex the parent is indeed updated, and in theory DocumentViewBox will receive the new prop. But there is nothing to trigger App component to update: its state does not change, it has no props, and it is not using context. So Overlay doesn't update either, meaning it continues to render its now outdated "children" with the old prop values.

We cannot simply lift the state and the DocumentViewBox JSX into App, because the documents are fetched in the level/start request which is made by MainComponent. This kind of limitation has come up before - see `mainBodyKey` state in MainComponent - and is another example of how storing level state in MainComponent with a useState hook, and passing it all down via prop drilling, is making life uncomfortable. We might be better off switching to a context-based store instead, for level state.

frontend/src/service/backendService.ts
```diff
@@ -1,16 +1,51 @@
...
+	return new URL(path, backendUrl());
 }
 
-async function sendRequest(path: string, options: RequestInit) {
-	return fetch(makeUrl(path), { ...options, credentials: 'include' });
+type RequestOptions<T> = Pick<RequestInit, 'method' | 'signal'> & {
+	body?: T;
+};
+
+async function get(
```
chriswilty, line 16: This is all starting to look like [axios](https://axios-http.com/)! But hey, our implementation here is really lightweight and focused.

frontend/src/components/DocumentViewer/DocumentViewBox.tsx
pmarsh-scottlogic: Always teaching!
well explained

frontend/src/service/documentService.ts
```diff
@@ -2,21 +2,20 @@ import { IDocument } from '@cyntler/react-doc-viewer';
...
-	});
+function processDocumentMetadata(docs: DocumentMeta[]) {
+	const baseUrl = backendUrl();
+	return docs.map(
+		({ fileName, fileType, folder }) =>
+			({
+				fileName,
+				fileType,
+				uri: `${baseUrl}${PATH}/${folder}/${fileName}`,
+			}) as IDocument
```
pmarsh-scottlogic, line 17: nice that we're using the proper types given by the api

pmarsh-scottlogic (approved):
Looks great! Lovely bit of optimisation, and well explained.

Didn't scour it with a magnifiying glass (or run it, I'm afraid) but the code all looks sensible.

Issue #917, opened 2024-07-15
Release 1.0.3


chriswilty:
Patch release:
- Document Viewer issues addressed
- Available GPT models updated to latest, including gpt-4o and gpt-4-turbo
- Package vulnerabilities resolved, and dependencies upgraded to latest minor, across all modules

PR #918, opened 2024-07-15
Prepare for release 1.0.3

chriswilty:
## Description

- Dependencies upgraded, vulnerabilities resolved
- Our own app version in all package.jsons set to 1.0.3

Part of #917 

## Checklist

Have you done the following?

- [ ] Added tests


backend/src/controller/defenceController.ts
```diff
@@ -254,14 +254,16 @@ function handleResetDefenceConfigItem(
 		currentDefences
 	);
 
-	const updatedDefences = req.session.levelState[level].defences;
-	if (updatedDefences === undefined) {
+	const updatedDefence = req.session.levelState[level].defences.find(
+		(defence) => defence.id === defenceId
+	);
+	if (!updatedDefence) {
```
chriswilty, line 260: Change here is a result of eslint pointing out that `updatedDefences` in original code could not be undefined.

frontend/src/components/DocumentViewer/DocumentViewBox.test.tsx
```diff
@@ -50,7 +49,7 @@ describe('DocumentViewBox component tests', () => {
 
 	const mockDocumentViewer = vi.hoisted(() => vi.fn());
 	vi.mock('@cyntler/react-doc-viewer', () => ({
-		default: (props: DocViewerProps) => {
```
chriswilty, line 53: DocViewerProps is no longer available as an export in the latest release, but in any case, in the cloud feature branch I've changed the test to render the actual DocViewer component, instead of mocking it, so soon enough this code will be removed.

PR #927, opened 2024-11-04
714: Show error is we fail to reset defences

### LINKED ISSUE #714: Reset defence configurations still shows a message saying 'defence configured successfully' - Just disable the button. Show error message on timeout instead.

calummcgowan:
## Description

- When running the frontend only, attempting to 'reset' any of the defence configurations will result in an error message being displaying in the chat. 
- Fixes #714

## Screenshots

Screenshot 2024-11-04 110448

## Notes

- Changes made to MainComponent.tsx: add chat message when resetDefence fails
- Changes to defenceService.ts: return null when DefenceResetResponse promise fails

## Concerns

- No unit tests added - may be irrelevant due to running only frontend 

## Checklist

Have you done the following?

- [] Added tests

### DIFF
--- a/frontend/src/service/defenceService.ts
+++ b/frontend/src/service/defenceService.ts
@@ -85,7 +85,7 @@ async function resetDefenceConfigItem(
 			return (await response.json()) as DefenceResetResponse;
 		},
 		async () => {
-			return null;
+			return await null;
 		}
 	);
 }
--- a/frontend/src/service/defenceService.ts
+++ b/frontend/src/service/defenceService.ts
@@ -75,13 +75,19 @@ async function resetDefenceConfigItem(
 	defenceId: DEFENCE_ID,
 	configItemId: DEFENCE_CONFIG_ITEM_ID,
 	level: LEVEL_NAMES
-): Promise<DefenceResetResponse> {
-	const response = await post(`${PATH}/resetConfig`, {
+): Promise<DefenceResetResponse | null> {
+	return post(`${PATH}/resetConfig`, {
 		defenceId,
 		configItemId,
 		level,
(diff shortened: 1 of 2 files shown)


frontend/src/service/defenceService.ts
```diff
@@ -75,13 +75,19 @@ async function resetDefenceConfigItem(
...
-): Promise<DefenceResetResponse> {
-	const response = await post(`${PATH}/resetConfig`, {
+): Promise<DefenceResetResponse | null> {
+	return post(`${PATH}/resetConfig`, {
 		defenceId,
 		configItemId,
 		level,
-	});
-	return (await response.json()) as DefenceResetResponse;
+	}).then(
```
chriswilty, line 83: We have a mix of promises and async/awaits here. Prefer async/await wherever possible (and here it is possible).

Note that [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch) does not throw an error for non-200 status codes, so this will fail when the service (legitimately) returns text/plain error messages, as the body then is not JSON. I appreciate that a) that was already happening, and b) the acceptance criteria specified only what should happen when the server is unavailable (which does indeed result in an error thrown) but we have the opportunity to make this more robust if we check for response.ok ...

As this project is so infrequently updated, may I ask for a little more from you? Firstly, use async/await and try/catch for consistency, and secondly, could you check for `response.ok` and if false, then console.warn the response text (and then return null)?

chriswilty (changes requested):
Sorry it took a while to get around to a review.

This does the job, thank you! I have a small request for a minor enhancement, and another to use async/await instead of a promise chain.

